PR #26, opened 2020-07-11
Add a `dstask rm/remove` command

dontlaugh:
We add two constants, CMD_RM and CMD_REMOVE, so user can type "rm" or
"remove", respectively.

### DIFF
--- a/cmd/dstask.go
+++ b/cmd/dstask.go
@@ -393,6 +393,7 @@ func main() {
 			"",
 			dstask.CMD_NEXT,
 			dstask.CMD_ADD,
+			dstask.CMD_REMOVE,
 			dstask.CMD_LOG,
 			dstask.CMD_START,
 			dstask.CMD_STOP,
--- a/cmd/dstask.go
+++ b/cmd/dstask.go
@@ -64,6 +64,23 @@ func main() {
 			dstask.MustGitCommit("Added %s", task)
 		}
 
+	case dstask.CMD_RM, dstask.CMD_REMOVE:
+		if len(cmdLine.IDs) < 1 {
+			dstask.ExitFail("%s", "missing argument: id")
+		}
+		ts := dstask.LoadTasksFromDisk(dstask.NON_RESOLVED_STATUSES)
+		for _, id := range cmdLine.IDs {
+			task := ts.MustGetByID(id)
+
+			// Mark our task for deletion
+			task.Deleted = true
+
+			// MustUpdateTask validates and normalises our task object
+			ts.MustUpdateTask(task)
+			ts.SavePendingChanges()
+			dstask.MustGitCommit("Removed: %s", task)
+		}
+
 	case dstask.CMD_LOG:
 		ts := dstask.LoadTasksFromDisk(dstask.NON_RESOLVED_STATUSES)
 
--- a/const.go
+++ b/const.go
@@ -32,6 +32,8 @@ const (
 
 	CMD_NEXT             = "next"
 	CMD_ADD              = "add"
+	CMD_RM               = "rm"
+	CMD_REMOVE           = "remove"
 	CMD_LOG              = "log"
 	CMD_START            = "start"
 	CMD_NOTE             = "note"
@@ -128,6 +130,8 @@ var NON_RESOLVED_STATUSES = []string{
 var ALL_CMDS = []string{
 	CMD_NEXT,
 	CMD_ADD,
+	CMD_RM,
+	CMD_REMOVE,
 	CMD_LOG,
 	CMD_START,
 	CMD_NOTE,
--- a/git.go
+++ b/git.go
@@ -38,11 +38,6 @@ func MustGitCommit(format string, a ...interface{}) {
 	fmt.Printf("\033[0m")
 }
 
-// MustGitRemove runs a "git rm" on the provided file path, relative to GIT_REPO.
-func MustGitRemove(relativePath string) {
-	MustRunGitCmd("rm", relativePath)
-}
-
 // leave file as an empty string to return directory
 func MustGetRepoPath(directory, file string) string {
 	dir := path.Join(GIT_REPO, directory)
--- a/git.go
+++ b/git.go
@@ -62,6 +62,6 @@ func InitialiseRepo() {
 }
 
 func Sync() {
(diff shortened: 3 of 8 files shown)


const.go
```diff
@@ -32,6 +32,8 @@ const (
 
 	CMD_NEXT             = "next"
 	CMD_ADD              = "add"
+	CMD_RM               = "rm"
```
naggie, line 35: I'd rather not have aliases baked in, as the tab completion for dstask makes it a bit redundant. Though, `rm `is muscle memory for most people that use a shell; perhaps this command should be `rm ` only

cmd/dstask.go
```diff
@@ -63,6 +64,21 @@ func main() {
...
+			dstask.ExitFail("%s", "missing argument: id")
+		}
+		ts := dstask.LoadTasksFromDisk(dstask.NON_RESOLVED_STATUSES)
+		for _, id := range cmdLine.IDs {
+			task := ts.MustGetByID(id)
+			relativePath := filepath.Join(task.Status, task.UUID+".yml")
+			dstask.MustGitRemove(relativePath)
+			// NOTE: do we need to do anything with the TaskSet tasksByUUID map?
+			// *ts.tasksByUUID[task.UUID]
+			// NOTE: Do we need to interact with the TaskSet or the IdsMap database?
```
naggie, line 78: Technically this won't matter as the IDs are recomputed every time `SavePendingChanges()` is called.

However I think it is cleaner to keep all task set management within the taskset methods. Perhaps each task could have a boolean `delete` which is never serialised -- `task.save()` would then perform the deletion when called by `SavePendingChanges()`. The interface to this could be via `MustUpdateTask()` given a mutated task returned by your existing call to `MustGetByID`

const.go
naggie: Actually this is apparently fine as I've already done this for other commands. I guess I'd rather be consistent, let's leave remove/rm in.

cmd/dstask.go
dontlaugh: I'll rework this to encapsulate the functionality within the taskset. :+1:

ard0gg:
Should updating help.go to include CMD_RM, CMD_REMOVE be part of this PR?

dontlaugh:
Thanks for your feedback, and for guidance regarding the codebase. I believe this is ready for review. 

I have done some manual testing between two workstations. 

I have opted to document the remove command as "remove", as I believe it will read more nicely when folks run `dstask help`. And I think most users will expect and discover the alias to "rm" on their own.

dontlaugh:
@naggie I've reworked this.
* mutation happens in the RM case statement, where a new Deleted field is set
* mutation happens in MustUpdateTask, as before, but if a task is deleted WritePending is not set

dontlaugh:
Ahhh! I ddin't mean to close!

help.go
```diff
@@ -182,6 +190,7 @@ Available commands:
 
 next              : Show most important tasks (priority, creation date -- truncated and default)
 add               : Add a task
+remove            : Remove a task
```
ard0gg, line 193: Should probably add this line to the README.md as well.

ard0gg:
Also need to update the CMD_COMPLETIONS case in dstask.go to include CMD_RM and CMD_REMOVE functions.

git.go
```diff
@@ -38,6 +38,11 @@ func MustGitCommit(format string, a ...interface{}) {
 	fmt.Printf("\033[0m")
 }
 
+// MustGitRemove runs a "git rm" on the provided file path, relative to GIT_REPO.
+func MustGitRemove(relativePath string) {
+	MustRunGitCmd("rm", relativePath)
```
naggie, line 43: not sure adding a function wrapper adds anything, probably best to call `MustRunGitCmd("rm", relativePath)` on its own

git.go
```diff
@@ -62,6 +67,6 @@ func InitialiseRepo() {
 }
 
 func Sync() {
-	MustRunGitCmd("pull", "--no-edit", "--commit", "origin", "master")
+	MustRunGitCmd("pull", "--no-rebase", "--no-edit", "--commit", "origin", "master")
```
naggie, line 70: Good idea, unrelated but that's fine and useful.

taskset.go
```diff
@@ -312,6 +315,11 @@ func (ts *TaskSet) SavePendingChanges() {
 			task.SaveToDisk()
 		}
 
+		if task.Deleted {
+			relativePath := filepath.Join(task.Status, task.UUID+".yml")
```
naggie, line 319: I think the task struct needs a `Delete()` method to be consistent with the path resolution and abstraction afforded by `SaveToDisk()` which does `filepath := MustGetRepoPath(t.Status, t.UUID+".yml")`

Note that giving git an absolute filepath the delete is fine.

taskset.go
```diff
@@ -185,7 +186,9 @@ func (ts *TaskSet) MustUpdateTask(task Task) {
 		task.Resolved = time.Now()
 	}
 
-	task.WritePending = true
+	if !task.Deleted {
```
naggie, line 189: I think it's best to decide whether to call `SaveToDisk()` in `SavePendingChanges()` rather than doing this early, as in theory the task could be mutated again before the SavePendingChanges.

taskset.go
naggie: On second thoughts, I consider deletion to be part of the job of `SaveToDisk()` as that already has a mechanism to delete all matching tasks in all directories (important if there are manual changes to the repository which could result in duplicate tasks.

naggie:
Looking good @dontlaugh! Just had a few more comments.

taskset.go
dontlaugh: So, no special handling of deleted tasks in `MustUpdateTask`, then?

dontlaugh:
Now we check for Deleted in `SaveToDisk`, and simply remove the file if the task is marked for deletion. I've removed the call to `git rm` altogether. A regular file delete will suffice, since we immediate stage and commit the deletion.

naggie:
Looks good -- happy to merge if you are

dontlaugh:
@naggie Sounds good. Thanks for the review!

naggie:
Merged -- thanks for your contribution @dontlaugh !

Issue #25, opened 2020-07-08
View markdown notes without editing


dontlaugh:
Sometimes it is nice to be able to print the markdown notes to a file like this:

```
dstask note 4 > output.md
```

Looking at the code here, it doesn't seem to be possible yet https://github.com/naggie/dstask/blob/master/cmd/dstask.go#L224

`dstask` could check if stdout is a tty, and if it is not, simply write the markdown to stdout. Otherwise, if the output is a tty, open a copy of the note in the $EDITOR (the current behavior).

naggie:
Actually I had planned that exact thing, except the issue is tracked on my local dstask haha. I should probably transfer them all to github.

I'll get round to it soon as I think it's a useful feature.

dontlaugh:
#27 should have fixed this. Reopen if there's a bug.

PR #27, opened 2020-07-11
Change to note command: print to stdout if it is not a TTY

dontlaugh:


dontlaugh:
This should fix #25

cmd/dstask.go
```diff
@@ -214,25 +214,32 @@ func main() {
 			dstask.MustGitCommit("Edited %s", task)
 		}
 
-	case dstask.CMD_NOTES:
-		fallthrough
-	case dstask.CMD_NOTE:
+	case dstask.CMD_NOTE, dstask.CMD_NOTES:
```
naggie, line 217: Hm, seems I have added aliases before! The `rm`/`remove` thing is fine by precedent, for now then.

naggie:
All good, thanks!

Issue #29, opened 2020-07-14
Add copy command


ard0gg:
I would like to create some template tasks that have some some generic process steps in the notes. 
This would be used for complex tasks that might be repeated for different projects (like a release checklist)

Assuming my template task were numbered 16, I imagine the command would look something like: 
`dstask copy 16 This is the description of the new task copied from 16 `

ard0gg:
Perhaps this command would be better:
`dstask 16 copy ....`

dontlaugh:
I like the motivation here, however there are some subtle complexities to a `copy` command. Specifically, projects, priorities, tags. Should these things be copied, as well? What is right for one user might not be right for another. This is where I would, personally, write a script that suits my needs.

But we have a problem. Even if we do not make a dedicated `copy` command, as it stands the default `task N` command will complain about stdout not being a terminal. This is similar to the issue in #25 

Therefore, it is **not yet** possible to write a copy function like this, which could be a version of what you want:

```sh
copytask () {
  local taskid=$1
  summary=$(dstask $taskid | awk '/^Summary/ { $1=""; $0=$0; } NF=NF')
  task add $summary
}
```
We should, at the very least, enable a scripting use case. Scriptability was what enabled taskwarrior to develop a rich and interesting ecosystem.

I will defer to naggie on what should or should not be in core.

ard0gg:
A script could work just as well. 
I would expect that and projects, priorities, tags would be copied by default, but additional modifiers in the command line would modify the copied task as necessary. I would expect that it would be sort of  CMD_ADD followed by a CMD_MODIFY. 

Excuse my coding. I'm not a strong programmer. So there's some psuedo-code 

```golang
case dstask.CMD_COPY:

                // add some logic to parse cmdLine.IDs and verify just one and that it's valid
             

		originalTask := dstask.LoadTasksFromDisk(dstask.NON_RESOLVED_STATUSES).mustgetbyid(id)

		context.PrintContextDescription()  
		cmdLine.MergeContext(context)
		task := dstask.Task{
			WritePending: true,
			Status:       dstask.STATUS_PENDING,
			Summary:      originalTask.Text+" - COPY" ,  //default summary text to append "- COPY"
			Tags:         originalTask.Tags,
			Project:      originalTask.Project,
			Priority:     originalTask.Priority,
			Notes:        originalTask.Note,
		}
			task = ts.LoadTask(task)
                        
                        // modify cmdline.IDs to remove originalTaskID and replace with task ID
[...]

PR #30, opened 2020-07-16
Added template/copy functionality

ard0gg:
Edit: Addresses #29 
I've decided to take a stab at this and make a contribution. 
I've added two functions:
- **template**
Use: dtask template ... or dtask \<id\> template
will either create a template if no ID is given or convert a task into a template. Templates have similar data model to recurring tasks, so are stored in the templates/ folder. 
- **show-templates**
Use: dstask show-template
 this shows current templates

Modified the add command to allow for the template:\<ID\> modifer. The way I've written it, the ID does not have to be a template. So this could function as a way to copy tasks that aren't necessarily templates. However, its intended purpose is for copying templates from the template/ directory as well as adding tags/priority/project to the added task from the add command. (needs some testing)

TODO: 

- Probably some cleanup in the const.go and dstask.go file. I didn't want to alter NON_RESOLVED_STATUS since that is used for a lot of the SHOW-* functions even though it makes sense to include STATUS_TEMPLATES into this array.  Looking for some input here as to to whether to alter NON_RESOLVED_STATUS to include templates and use a different array name for displaying non-template, non-resolved tasks or to create a different array with NON_RESOLVED_STATUS + STATUS_TEMPLATES.

- Need to update the Edit command to allow for updating template tasks

- Update help.go with new functions.  

- Edit: The add template:<id> paradigm currently does not address Subtasks or Dependencies within the template task.

- You tell me


ard0gg:
#29

cmd/dstask.go
```diff
@@ -45,9 +45,29 @@ func main() {
 		ts.DisplayCriticalTaskWarning()
 
 	case dstask.CMD_ADD:
-		ts := dstask.LoadTasksFromDisk(dstask.NON_RESOLVED_STATUSES)
+		ts := dstask.LoadTasksFromDisk(append(dstask.NON_RESOLVED_STATUSES, dstask.STATUS_TEMPLATES))
 
-		if cmdLine.Text != "" {
+		if cmdLine.Template != 0 {
```
ard0gg, line 50:
```go 
if cmdLine.Template > 0
```

cmdline.go
```diff
@@ -52,6 +53,10 @@ func (cmdLine CmdLine) String() string {
 		args = append(args, cmdLine.Priority)
 	}
 
+	if cmdLine.Template != 0 {
```
ard0gg, line 56:
``` go
 if cmdLine.Template > 0 {
```

naggie:
Thanks for the PR, I agree with your reasoning about allowing copying with the same syntax -- it's safe as IDs are unique across statuses and is convenient. Though, perhaps dstask should tell the user about real templates when they use the template specifier on a regular task.

> Probably some cleanup in the const.go and dstask.go file. I didn't want to alter NON_RESOLVED_STATUS since that is used for a lot of the SHOW-* functions even though it makes sense to include STATUS_TEMPLATES into this array. Looking for some input here as to to whether to alter NON_RESOLVED_STATUS to include templates and use a different array name for displaying non-template, non-resolved tasks or to create a different array with NON_RESOLVED_STATUS + STATUS_TEMPLATES.

Template tasks should definitely be part of NON_RESOLVED_STATUSes, we should simply not display template tasks on the requisite views. There are 3 reasons:

* The original and standing purpose of NON_RESOLVED_STATUS is to just avoid loading resolved tasks, which over time tend to be numerous compared to other tasks. I have thousands -- it's just a performance thing.
* The ID system relies on consistently loading the same set of numerically addressable tasks every time in order to avoid conflicts
* The edit command would just work as below

> Need to update the Edit command to allow for updating template tasks

In theory it should just work, as template tasks are just regular tasks with a different status.

naggie:
Oh, another thing: https://github.com/naggie/dstask/blob/da0248fd270d59542fb63f9622fb1ed805bdb054/cmd/dstask.go#L348 will have to be updated so the bash/zsh completions work for `template`, `show-templates` etc

naggie:
Just tested this, happy to merge it if it's ready.

btw -- I noticed `template:` does not complete for the bash completions. I'll add that once this is in master unless you want to -- I think `template:` could expand to the available task numbers and should only be for the `add` command.

ard0gg:
Fell free to merge. I'll probably not have time to implement the "template:" bash completion this weekend.  If you aren't going to get to it before next weekend I can give a shot next week and issue another PR.

naggie:
Merged! Thanks for your contribution, @ard0gg !!

naggie:
btw, I pulled v0.19 to rename `STATUS_TEMPLATES=templates` to `STATUS_TEMPLATE=template` to be consistent, state should be singular. You'll have to rename your templates directory in your git repository. Sorry I didn't see this earlier, I'll add it to the release notes.

PR #31, opened 2020-07-18
Docs correction: remove command simply deletes and commits.

dontlaugh:
It does not actually use git rm.


dontlaugh:
Sorry, missed this yesterday. Small thing, but I expect folks will wonder how it works. :sweat_smile:

naggie:
Ok -- not sure the mechanics are important but I suppose the target audience is almost certainly technical!

PR #32, opened 2020-07-19
Improve template Auto-Completion

ard0gg:
Current implementation of "template:" auto-completion for the "add" command inserts a space after "template:"

This doesn't address that directly, but it does include the available tasks that can be copied. When tab-completion occurs, a space will not be inserted after "template:" Presently this will autocomplete for all tasks under NON_RESOLVED_STATUSES. However, removing the comment around the If statement will change this such that only Template statuses will autocomplete. 

Note: The order of tasks appear in increasing order by first digit, then second digit... rather than overall integer value. I'm not sure how to adjust this yet. I don't know if it's a function of dstask or of the completion bash script. 

```bash
@penguin:~/dstask-develop$ dstask add template:
1   10  11  12  13  14  15  16  17  18  19  2   20  3   4   5   6   7   8   9   
```

### DIFF
--- a/cmd/dstask.go
+++ b/cmd/dstask.go
@@ -492,6 +492,15 @@ func main() {
 				ts.Filter(context)
 			}
 
+			// templates
+			if cmdLine.Cmd == dstask.CMD_ADD {
+				for _, task := range ts.Tasks() {
+					if task.Status == dstask.STATUS_TEMPLATE {
+						completions = append(completions, "template:"+strconv.Itoa(task.ID))
+					}
+				}
+			}
+
 			// priorities
 			completions = append(completions, dstask.PRIORITY_CRITICAL)
 			completions = append(completions, dstask.PRIORITY_HIGH)
@@ -508,18 +517,6 @@ func main() {
 			for tag := range ts.GetTags() {
 				completions = append(completions, "+"+tag)
 				completions = append(completions, "-"+tag)
-			}
-		}
-
-		if cmdLine.Cmd == dstask.CMD_ADD {
-			ts := dstask.LoadTasksFromDisk(dstask.NON_RESOLVED_STATUSES)
-			// If statement below is uncommented, but no taskIDs are added to completions
-			// then tab completion will insert space after "template:"
-			//completions = append(completions, "template:")
-			for _, task := range ts.Tasks() {
-				// if task.Status == dstask.STATUS_TEMPLATE {
-				completions = append(completions, "template:"+strconv.Itoa(task.ID))
-				//}
 			}
 		}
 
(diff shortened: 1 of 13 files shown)


cmd/dstask.go
```diff
@@ -512,7 +512,15 @@ func main() {
 		}
 
 		if cmdLine.Cmd == dstask.CMD_ADD {
-			completions = append(completions, "template:")
+			ts := dstask.LoadTasksFromDisk(dstask.NON_RESOLVED_STATUSES)
+			// If statement below is uncommented, but no taskIDs are added to completions
+			// then tab completion will insert space after "template:"
+			//completions = append(completions, "template:")
```
ard0gg, line 518: As presently implemented "template:" will not autocomplete if there are no tasks within NON_RESOLVED_STATUSES. This make sense as there are no tasks to copy. 

However, if lines 520/522 is uncommented you may want to consider uncommenting this line as well. Otherwise "template:" will not autocomplete if there aren't any existing template tasks. However, note in this case that a space will be inserted after "template:". I'm not sure how to fix this exactly.

naggie:
Good point, the space is annoying. Completing to just template IDs is the right way (such that we'll get no completion if there are no numeric templates).

So I think 520/522  should be uncommented, and 516-518 deleted for that behaviour.

Finally, we should load the tasks from disk only once for the completions, as it's an expensive operation.


Thanks!

ard0gg:
Hi naggie,

Thanks for the review. I've gone ahead and make the changes such that it only completes for Template IDs. Also the logic for the addition of template completions was outside the if statement where LoadTaskFromDisk() was called. So I moved it inside. 

I believe it should be ready to merge unless you see any other improvements. 

Thanks!

cmd/dstask.go
```diff
@@ -231,7 +231,7 @@ func main() {
 
 	case dstask.CMD_CONTEXT:
 		if len(os.Args) < 3 {
-			fmt.Printf("Current context: %s", context)
+			fmt.Printf("Current context: %s\n", context)
```
ard0gg, line 234: Unrelated. 
Added a newline here. Calls of dstask context would result in 

``` bash
@ard0gg:~/dstask-develop/ard0gg$ dstask context
Current context: @ardogg:~/dstask-develop/ard0gg$
```

Now result in 
``` bash
@ard0gg:~/dstask-develop/ard0gg$ ./dstask context
Current context:
@ard0gg:~/dstask-develop/ard0gg$
```

naggie: Oh, thanks. I never noticed that as I have a multiline prompt

naggie:
Great, thanks @ard0gg it works very well.

Issue #28, opened 2020-07-12
Add single-character tags at the command line


dontlaugh:
It seems that single-character tags cannot be added at the command line. The can be added via `dstask edit` however.

This little shell session demonstrates the issue by attempting to add `+x` as a tag to an existing issue with `modify`.

Also note that `dstask add +x Try OpenBSD` yields a **new** task with "+x" in the task summary. 

```
❯ dstask context +servers
❯ dstask
Active context: +servers
Name      Value                                 
ID        13                                    
Priority  P2                                    
Summary   consolidate all servers               
Status    pending                               
Project                                         
Tags      servers                               
UUID      4ab9db58-3f3e-484e-ac91-21cf10ed51ff  
Created   2020-07-04 11:08:55.02537815 -0400 EDT

❯ dstask modify 13 +x

Modified 13: consolidate all servers
No changes detected

❯ dstask
Active context: +servers
Name      Value                                 
ID        13                                    
Priority  P2                                    
Summary   consolidate all servers               
Status    pending                               
Project                                         
Tags      servers                               
UUID      4ab9db58-3f3e-484e-ac91-21cf10ed51ff  
Created   2020-07-04 11:08:55.02537815 -0400 EDT

❯ dstask add +x +servers Try OpenBSD
Active context: +servers

Added 7: +x Try OpenBSD
[master b9476a0] Added 7: +x Try OpenBSD
 1 file changed, 12 insertions(+)
 create mode 100644 pending/2865b8af-5f49-47c6-a2ab-c0cc3799a044.yml

❯ dstask                            
Active context: +servers
ID  Pr  Tags       Summary                
13  P2  servers    consolidate all servers
7   P2  servers    +x Try OpenBSD    
```

Would you accept a contribution that fixes this and, perhaps adds a unit test or two around the parsing code?

naggie:
Definitely, thanks. Sorry I haven't had time to get round to reviewing your other PRs, busy weekend and working today. Should get time tonight.

dontlaugh:
Don't sweat it! I'm just logging issues as I go. :blush:

naggie:
Resolved in https://github.com/naggie/dstask/commit/279bc0da2a1840698868e90ae94cffee3f05c812

PR #33, opened 2020-07-20
Add cmdline_test.go and supporting libraries

dontlaugh:


dontlaugh:
Opening this as WIP, since there isn't an associated issue. I only briefly waved my hands at it in #28 

I do have a hunch that this project would benefit from unit tests, over the long term:
* We do string parsing, and as we add features that will get more complicated
* Implementing recurring tasks will require working with time, and that can be error prone even with Go's nice standard library
* Interest in this project will grow, and tests will help contributors keep their code quality high, since testable designs usually correlate with good designs

Of course, there are tradeoffs:
* it's undoubtedly more work
* it's verbose, especially in Go

It's up to you. What do you think? Shall we introduce unit tests? Or just enhance the integration test suite in the style of **smoketest.sh**?

naggie:
I think it's absolutely worth introducing unit tests, it's been on my dstask todo list for some time. I greatly appreciate the start you've made here.

As for the extra dependencies I'll trust your judgement and go with them.

I'll add this to CI as well.

Perhaps it's worth merging now so we can get the CI ball rolling?

Thanks!

dontlaugh:
@naggie Right on. Thanks!

Issue #18, opened 2020-05-10
Modifying resolved tasks


qTipTip:
Hey!

First of all, thanks for a great task-manager!

I have a question. Is it possible to edit already resolved tasks? I.e., for instance to add notes pertaining to the resolution of the task.

Running `dstask show-resolved` shows the list of resolved tasks, but I can't seem to find an ID I can reference in a call to `dstask modify`. Is this something I have just missed, or does this feature not exist?

Cheers in advance!

naggie:
Hi!

> First of all, thanks for a great task-manager!

You're welcome, I'm glad you like it!

> Is it possible to edit already resolved tasks? 

Not currently via cli, I've been meaning to allow manipulation of resolved tasks.

Resolved tasks no longer have a numeric ID, but still retain a UUID which remains the same throughout the lifetime of the task. I think the way I will do it is allow addressing of resolved task by their UUID, which would be listed in `show-resolved`.

I can make it such that you don't have to use the entire UUID, just enough to disambiguate. For instance, the resolved task with UUID `2cb8aa23-4d02-4ab8-8665-b154ed4a30ee` could be edited with `dstask edit 2cb8aa` or something. dstask could ask for more characters if that matches more than one task. `show-resolved` could list the first 6 by default where possible.

Does this sound like a good solution?

qTipTip:
Hi again!

UUID sounds very reasonable. This is what git uses as well, no?

What are the chances of two UUIDs matching?

stianlagstad:
You'll have to be veeeeeeeery lucky for that to happen, @qTipTip !

> For example, the number of random version-4 UUIDs which need to be generated in order to have a 50% probability of at least one collision is 2.71 quintillion ..
> This number is equivalent to generating 1 billion UUIDs per second for about 85 years.
ref https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions

naggie:
> UUID sounds very reasonable. This is what git uses as well, no?

Git uses a merkel tree using sha1 hashing. dstask uses UUID4s already for a globally unique task ID. As @stianlagstad says, collisions are.... improbable :)

@stianlagstad -- thanks for posting an article on dstask, by the way. I enjoyed reading it.

NerdsvilleCEO:
@naggie I can volunteer to update `dstask` to show uuid with the `show-resolved` command if this isn't already being worked on

ThreeFx:
[...]

PR #37, opened 2020-07-24
show-resolved shows UUID, can use edit/modify/notes command resolved tasks

ard0gg:
This is a work in progress. 
Edit: Addresses #18 
I've modified the show-resolved display to also include the first 8 digits of
the UUID in the first column. As shown below:

```
> Week 29, starting Tue 14 Jul 2020

UUID      Resolv  Pr  Tags  Project   Summary
a0a66fef  Tue 14  P2  work  12button  pull part numbers for design
a7813a92  Thu 16  P2  work  #####     release BOM/Scheamtic


> Week 30, starting Mon 20 Jul 2020

UUID      Resolv  Pr  Tags  Project   Summary
e4998b2b  Mon 20  P2  work  12button  pass schematic off to doug
caf99371  Mon 20  P2  work  12button  review schematic
9bd43844  Mon 20  P2  work  rgb       release Master schematic/BOM
5b04a0ac  Mon 20  P2  work  rgb       release slave schematic/BOM
39eba78c  Mon 20  P2  work  amp       perform troubleshoot update
5a86a39b  Mon 20  P1  work  12cpm     Finish DFMEA
7e32a1d4  Mon 20  P1  work  qhb       complete DFMEA
93ec556d  Tue 21  P2  work  qhb       update specification
931b58fc  Wed 22  P2  work  qhb       review DFMEA
9bc83c4d  Wed 22  P2  work  12cpm     review DFMEA
2eec2e8a  Thu 23  P1  work  12cpm     Release DFMEA
c3c825eb  Fri 24  P2  work  12cpm     modify d6 to match QHB
14 tasks.
Active context: +work
```

Additionally added the ability to edit resolved tasks by using the "uuid:"
modifier with the edit command. I'm not sure if this is the best command to use. Would it be better to use notes or modify? Or have the capability to use those commands in addition to the edit command? 

Right now only resolved tasks can be found using the "uuid:" tag. 

CMD_EDIT will check if the command line includes a UUID and load
resolved tasks if so. It will then check to see if it can match the user
entered UUID with any resolved tasks. If only one matches, it is edited.
User is prompted if no UUIDs are matched:
```
ard0gg:~/src/dstask/ard0gg$ task edit uuid:5adfaddfa
No Resolved Tasks Match UUID: 5adfaddfa!
```

 or more than 1 UUID matches. In the case that more
than 1 UUID matches, the matching UUIDs are listed. Right now the display of the tasks is 
hard coded in CMD_EDIT. Would like to be able to use filters/functions in display.go eventually. 
```
ard0gg:~/src/dstask/ard0gg$ task edit uuid:9
Multiple Tasks match UUID: 9

UUID                                    Summary
915cb9f5-0ee4-4824-a865-e98830f7c8f4    add switch for BL PWM
931b58fc-e0da-4de0-828f-5ab7485d8dc3    review DFMEA
93ec556d-ab8e-40c3-97e1-a4782813e288    update specification
9bc83c4d-f39d-4689-8609-87227da7bbcf    review DFMEA
9bd43844-9cdf-487d-8b39-b46c164a0122    release Master schematic/BOM

Provide provide more specific UUID
```

Completion of "uuid:" works for edit iff there are no IDs in the command line.  However, there is the issue where a space is presented after "uuid:".  I'm not intending to include completion of the task.UUIDs.


naggie:
This is how I was planning to implement it. Though, I had thought we could do it without the `uuid:` prefix; the parser would match on anything that looks like a UUID4 substring of more than (say) 4 characters with the listing disambiguation mechanism.

Of course this would produce some false positives sometimes (when searching for a string)  though I'd wager rarely, but the result may be benign.

cmd/dstask.go
```diff
@@ -265,34 +265,80 @@ func main() {
...
 
-			data, err := yaml.Marshal(&task)
-			if err != nil {
-				// TODO present error to user, specific error message is important
-				dstask.ExitFail("Failed to marshal task %s", task)
+				err = yaml.Unmarshal(data, &task)
+				if err != nil {
+					// TODO present error to user, specific error message is important
+					// TODO reattempt mechanism
+					dstask.ExitFail("Failed to unmarshal yml")
```
dontlaugh, line 304:
```suggestion
					dstask.ExitFail(fmt.Sprintf("Failed to unmarshal yml: %v", err))
```
If this error does happen, it would probably mean a corrupted file or a manual edit. This is one way we could show the underlying error.

taskset.go
```diff
@@ -229,6 +229,22 @@ func (ts *TaskSet) MustGetByID(id int) Task {
...
+	for _,task := range ts.tasks {
+		if strings.HasPrefix(task.UUID,uuid){
+			uuids = append(uuids, task.UUID)
+		}
+	}
+	return uuids
+}
+func (ts *TaskSet) MustGetByUUID(uuid string) Task {
+	if ts.tasksByUUID[uuid] == nil {
+		ExitFail("No task with UUID %v exists.", uuid)
```
dontlaugh, line 243: We should probably return `(Task, error)` here, and call `ExitFail` at a higher level.

util.go
```diff
@@ -44,6 +45,25 @@ func MustGetUUID4String() string {
 	return u.String()
 }
 
+func IsValidPartialUUID4String(str string) bool {
```
dontlaugh, line 48: I would add a **util_test.go** file, and a table-driven unit test for this function. Just a few success/failure cases would do.

dontlaugh:
On the question of using a uuid: prefix vs. automatic detection of uuid, I would vote for keeping the prefix.

1) if a prefix aids in writing the command line completion scripts, that seems like the bigger benefit
2) automatic detection of UUID could always be added later

cmd/dstask.go
naggie: Yes, that could be helpful. We could utilise [ConfirmOrAbort](https://github.com/naggie/dstask/blob/master/util.go#L20) in a loop to enable the user to correct mistakes, too (another one I had planned but never got round to)

taskset.go
naggie: For testability, yes I agree. I tried to reduce lines + error handling with `ExitAbort` but it's not practical when adding unit tests when called deep within.

naggie:
> On the question of using a uuid: prefix vs. automatic detection of uuid, I would vote for keeping the prefix.
> 
>     1. if a prefix aids in writing the command line completion scripts, that seems like the bigger benefit
> 
>     2. automatic detection of UUID could always be added later


Ok, I think you've persuaded me with the command line completion point. Though, it will be slow if we're completing for resolved tasks (probably the primary use case for addressing by UUID). 

Caching resolved UUIDs might be necessary but I'd like to avoid that unless there's no choice. Once the OS caches the files, loading all resolved tasks is fast enough so far.

cmdline.go
```diff
@@ -95,8 +101,13 @@ func ParseCmdLine(args ...string) CmdLine {
 		}
 
 		if s, err := strconv.ParseInt(item, 10, 64); !IDsExhausted && err == nil {
-			ids = append(ids, int(s))
-			continue
+			if int(s) <= MAX_TASKS_OPEN && string(lcItem[0]) != "0" {
```
ard0gg, line 104: Added a little more error checking here. I ran into the scenario where my first 7-8 digits of the UUID were all numeric and they resolved to IDs rather than UUIDs.

I also added the condition that if the first digit of lcItem is 0, it cannot be parsed as an ID. I ran into the scenario where the first 6 dgits of my UUID were all numeric with a leading 0.

With MAX_TASKS_OPEN set to 10000, there is a ~0.08% chance of conflict between UUID and ID given the first 6 digits.
Edit: Although if the requirement that the UUID needs to be greater than 4 digits, the only potential conflict (assuming MAX_TASKS_OPEN = 10000) is the case that ID == 10000, or 1 in 15^5. There is still the case that you might run into issues while while trying to match other text

ard0gg:
> Ok, I think you've persuaded me with the command line completion point. Though, it will be slow if we're completing for resolved tasks (probably the primary use case for addressing by UUID).

So are we thinking supporting UUID only with a "uuid:"  tag or would we still like the parser to try and match anything that looks like a UUID4 substring as well?

ard0gg:
Thank you @dontlaugh and @naggie  for the review. I really appreciate the feedback. I will work on incorporating the suggestions you've made.

ard0gg:
Another question I had.... Do we want to allow for status of resolved tasks to be changed?

naggie:
> So are we thinking supporting UUID only with a "uuid:" tag or would we still like the parser to try and match anything that looks like a UUID4 substring as well?

I think we should go with your idea of just having the `uuid:` prefix having thought about it a bit. I think it's natural syntax, given you can query with `project:` too.

> Thank you @dontlaugh and @naggie for the review. I really appreciate the feedback. I will work on incorporating the suggestions you've made.

No problem, thanks for putting the effort in. I'm really pleased to see contributors + users!

> Another question I had.... Do we want to allow for status of resolved tasks to be changed?

I think definitely --  it could be useful and we can just add it to the valid status transition list: https://github.com/naggie/dstask/blob/2cee8103519c1fd1a4d0e25d263be7689241c1e8/const.go#L115 (consumed in util.go by IsValidTransition)

taskset.go
```diff
@@ -228,20 +230,55 @@ func (ts *TaskSet) MustGetByID(id int) Task {
...
-		ExitFail("No task with UUID %v exists.", uuid)
+}*/
+
+func (ts *TaskSet) MustGetByUUID(uuid string) (Task, error) {
+	if !IsValidUUID4String(uuid) {
+		errMsg := fmt.Sprintf("UUID: %s is not a valid UUID4 string", uuid)
+		return Task{}, errors.New(errMsg)
+	} else if ts.tasksByUUID[uuid] == nil {
+		errMsg := fmt.Sprintf("No task with UUID %v exists in %v.", uuid, ts)
+		return Task{}, errors.New(errMsg)
```
dontlaugh, line 278: You can use the `fmt.Errorf` function, which takes formatting strings and returns an error at the same time.
https://golang.org/pkg/fmt/#Errorf

ard0gg: Awesome. Thanks

ard0gg:
Looking for thoughts on resolving partial UUID4 strings. I now have completions working for the commands Edit/Modify/Note when an ID is not present in the CmdLine. Do we want to rely on completions to fill in whole UUID4 strings?

naggie:
I think that UUIDs are only likely to be used to reference resolved tasks, given the convenience of numeric IDs as well as the fact that no current non-resolved view shows the UUIDs. Given the performance implication of autocompleting non-resolved tasks, I suggest we just don't do it.

The disambiguation system that shows the matching prefixes is a good compromise in my opinion.

taskset.go
```diff
@@ -220,14 +221,69 @@ func (ts *TaskSet) FilterUnorganised() {
...
+		if err != nil {
+			return Task{}, err
+		}
+		return *ts.tasksByUUID[uuid], nil
+	} else if ts.tasksByUUID[uuid] == nil {
+		return Task{}, fmt.Errorf("No task with UUID %v exists in %v.", uuid, ts)
+	} else {
+		return *ts.tasksByUUID[uuid], nil
+	}
+}
```
ard0gg, line 273: After reviewing this yesterday. I'm not sure that I'm happy with how I've implemented this. Line 263 calls SearchForUUID, but I don't think that function is necessary. I could use the Filter method of TaskSet. For example, by adding the following to the `func (task *Task) MatchesFilter(cmdLine CmdLine)` I can filter on UUIDs. 

```go 
if cmdLine.UUID != "" && !strings.HasPrefix(task.UUID, uuid) {
    return false
}
```

I wonder if it wouldn't be better to pass a CmdLine to MustGetByUUID() and do something like this...
```go
if !IsValidUUID4String(cmdLine.UUID){
    if !IsValidPartialUUID4String(cmdLine.UUID) {
        // return Task{}, error
    } else {
        ts.Filter(cmdLine)
        tasks := ts.Tasks()
        if len(tasks) == 0 {
            // return Task{}, error
        } else if len(tasks) > 1 {
            ts.ShowMatchingUUIDs  // function to be created in display.go
            // return Task{}, error
        } else {
            return tasks[0], nil
        }
    }
...
```
There may be a tradeoff in performance... Putting the UUID test at the bottom of MatchesFilter() means that when searching for a resolved UUID, every resolved task will first have to check for matching ID, tags, antitag, antiprojects, projects, priority and Summary/Note text before checkign for matching UUID. Putting it more towards the top just means that when this is called for TaskSets loaded from NON_RESOLVED_STATUSES, they will be checking for cmdLine.UUIDs when there will be none. Maybe that's ok. 

 Using ts.Filter(cmdLine) has the added benefit of being able to filter on things other than just UUID (tags, projects, summary/notes text, etc). I can see this coming in handy when working with resolved tasks in the thousands. I have already seen some of my resolved UUIDs match up 6-7 digits with just a few hundred tasks. Just adding a tag or a project name to the cmdline may remove the need for disambiguation most of the time. 

 I also ran into the issue within SearchForUUID() where I couldn't create the disambiguation display code in display.go easily because I wasn't working on the TaskSet ts itself. Most of the display functions rely on the task.filtered parameter.

naggie:
> I think that UUIDs are only likely to be used to reference resolved tasks, given the convenience of numeric IDs as well as the fact that no current non-resolved view shows the UUIDs. Given the performance implication of autocompleting non-resolved tasks, I suggest we just don't do it.
> 
> The disambiguation system that shows the matching prefixes is a good compromise in my opinion.

Oops, I meant _resolved_ tasks -- completing non-resolved tasks is fine as the number of non-resolved tasks is generally stable -- resolved tasks keep adding up forever, and aren't generally loaded into the OS disk cache as they're only loaded for various reports.

Regarding processing UUIDs: generally speaking, once tasks are loaded iterating over the tasks to aggregate things isn't expensive. I deliberately haven't attempted to optimise aside from not loading resolved tasks all the time.

----

I'll leave you both to the discussion and implementation if that's OK (considering the effort and thought going into it, thanks) as we've just had our second baby a few days ago :)

dontlaugh:
CONGRATS!!!! :blush:

naggie:
Thanks :)

ard0gg:
Congrats on the new baby, Naggie! 

Thank you both for taking the time to review. Sorry I've also not had much free time lately, but I plan to take a closer look this weekend.

naggie:
Thanks @ard0gg :-)

No need to apologise, or any pressure :+1:

cgardner:
It seems like there are some conflicts in these changes now.

PR #40, opened 2020-08-02
Minor cleanup in func main

dontlaugh:
- Remove duplicate code for `dstask.CMD_SHOW_OPEN`
- Print error if yaml.Unmarshal fails


dontlaugh:
Oh i completely goofed. These are _not_ the same :sweat:

Issue #41, opened 2020-08-02
TaskSet and func main refactor


dontlaugh:
A few times now we've hinted at a desire to refactor `func main`. This will involve `TaskSet` as well, since manipulating this type is one of main's big jobs right now.

We can talk about some possible designs here. We'll want something testable and extensible, if possible. For instance, how might we accommodate #12 and/or #16 ? We don't want to come up with something that makes those awkward to implement.

dontlaugh:
An analysis of our current commands yields the following

### Commands that update state on disk

* add
* remove
* template
* log
* start
* stop
* done, resolve
* context
* modify
* edit
* note
* undo
* git
* import-tw

### Commands without side effects
Note that the sync command does change the state of the world, technically. But, from the perspective of the local task database, there are no changes.

* sync
* show-active
* show-paused
* show-open
* next (also the default command)
* show-projects
* show-tags
* show-templates
* show-resolved
* show-unorganised
* help
* version

dontlaugh:
We might consider a separation of responsibilities like this:

`func main()`
* reads configuration from the environment (i kind of include 'Context' in this category of runtime config)
* parses command line to determine which command is being run
* calls that command routine
  * commands might be utilities, such as completion script generators
* returns an exit code to the shell

`TaskSet`
* Constructor reads a subset of tasks from disk (depending on options, such as "status")
* This subset of tasks is held in memory, and we either
  a) make changes: add new task, start/stop, add notes, persist those changes to disk, etc, then render sensible output
  b) make no changes, construct a table for rendering, render the output

Changes:
* set context on TaskSet so it can control rendering of current context (usually is printed before the task table)
* add a `NewTaskSet` constructor that takes more options than `LoadTasksFromDisk`. We could use the functional options pattern, to allow accepting a variadic list of options.
```go
func NewTaskSet(repoPath string, opts ...TaskSetOpt) (*TaskSet, error) {
  var tso taskSetOpts
  for _, opt := range opts {
    opt(&tso)
  }
  // if tso.This or tso.That do stuff
}

type TaskSetOpt func(opts *taskSetOpts)

func WithStatuses(statuses ...string) TaskSetOpt {
  return func(opts *taskSetOpts) {
    opts.statuses = append(opts.statuses, statuses...)
  }
}

type taskSetOpts struct {
  statuses  []string
  renderStrategy string
}

// ResolveTasks gets a list of tasks by id, marks them resolved, and commits the result.
[...]

PR #43, opened 2020-08-22
Refactor func main, new TaskSet constructor

dontlaugh:
A new TaskSet constructor that takes functional options. The idea is that since dstask is a one-shot command line
program, we can centralize sorting, filtering, loading of tasks in the constructor of TaskSet.

Only status filtering options `WithStatuses` and `WithoutStatuses` are included in this PR. Other method-based
sorting mechanisms are retained to keep this from getting too big.

Global variables `GIT_REPO`, `STATE_FILE`, and `IDS_FILE` were removed. These are now initialized on program
startup, where we can apply overrides from configuration (such as environment variables). These are combined into a more generic `Config` type, passed down to command functions that need them, and passed down again to lower-level functions that need them. 

Refs #41

### DIFF
--- a/cmd/dstask.go
+++ b/cmd/dstask.go
@@ -70,7 +70,7 @@ func main() {
 		}
 
 	case dstask.CMD_CONTEXT:
-		if err := dstask.CommandContext(repoPath, idsFilePath, stateFilePath, state, ctx, cmdLine); err != nil {
+		if err := dstask.CommandContext(stateFilePath, state, ctx, cmdLine); err != nil {
 			dstask.ExitFail(err.Error())
 		}
 
@@ -95,7 +95,7 @@ func main() {
 		}
 
 	case dstask.CMD_SYNC:
-		if err := dstask.CommandSync(repoPath, idsFilePath, stateFilePath); err != nil {
+		if err := dstask.CommandSync(repoPath); err != nil {
 			dstask.ExitFail(err.Error())
 		}
 
--- a/cmd/dstask.go
+++ b/cmd/dstask.go
@@ -44,20 +44,8 @@ func main() {
 		}
 
 	case dstask.CMD_RM, dstask.CMD_REMOVE:
-		if len(cmdLine.IDs) < 1 {
-			dstask.ExitFail("%s", "missing argument: id")
-		}
-		ts := dstask.LoadTasksFromDisk(dstask.NON_RESOLVED_STATUSES)
-		for _, id := range cmdLine.IDs {
-			task := ts.MustGetByID(id)
-
-			// Mark our task for deletion
-			task.Deleted = true
-
-			// MustUpdateTask validates and normalises our task object
-			ts.MustUpdateTask(task)
-			ts.SavePendingChanges()
-			dstask.MustGitCommit("Removed: %s", task)
+		if err := dstask.CommandRemove(repoPath, context, cmdLine); err != nil {
+			dstask.ExitFail(err.Error())
 		}
 
 	case dstask.CMD_TEMPLATE:
--- a/cmd/dstask.go
+++ b/cmd/dstask.go
@@ -68,16 +68,8 @@ func main() {
 		}
 
 	case dstask.CMD_DONE, dstask.CMD_RESOLVE:
-		ts := dstask.LoadTasksFromDisk(dstask.NON_RESOLVED_STATUSES)
-		for _, id := range cmdLine.IDs {
-			task := ts.MustGetByID(id)
-			task.Status = dstask.STATUS_RESOLVED
-			if cmdLine.Text != "" {
-				task.Notes += "\n" + cmdLine.Text
-			}
-			ts.MustUpdateTask(task)
-			ts.SavePendingChanges()
-			dstask.MustGitCommit("Resolved %s", task)
+		if err := dstask.CommandDone(repoPath, context, cmdLine); err != nil {
+			dstask.ExitFail(err.Error())
 		}
 
 	case dstask.CMD_CONTEXT:
--- a/cmd/dstask.go
+++ b/cmd/dstask.go
@@ -5,7 +5,6 @@ import (
 	"os"
 	"strconv"
 	"strings"
-	"time"
 
 	"github.com/mvdan/xurls"
 	"github.com/naggie/dstask"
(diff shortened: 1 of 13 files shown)


dontlaugh:
@naggie This is still a WIP because I am getting the same test failure from *smoketest.sh* as the CI system. I've moved out all the commands to **commands.go**, except for the completion script generator, which is in **completions.go**.

Once I get the test passing, I'll ask for a review, since this PR is quite large as it is.

There is still more I'd like to integrate into `NewTaskSet`, such as filter options and sorting options. But that can come later.

naggie:
Ok sounds good! Thanks.

dontlaugh:
I'm getting closer on this, and the tests are passing. I do think there is more cleanup and simplification to do here, but it's ready for folks to take a look.

One thing we hadn't talked about before: I removed a few global variables, and that's meant passing more arguments in many files. I did this in the context of trying to sort out the failing smoke test. In retrospect, this wasn't 100% necessary, but I'd rather not back out the change.

dontlaugh:
@naggie @ard0gg Take a look when you get a chance.

naggie:
Generally looks good after a quick run through. I'll look in more detail tonight and I'll be using a binary from this branch for the next 2 days. Looks as though you've put a lot of effort into this, so thanks!

I've found 1 regression so far: `dstask -- show-resolved` doesn't correctly ignore the context (though `dstask show-resolved --` does). It seems the former is interpreted as the default `next` command searching for the string `show-resolved`.

cmd/dstask.go
```diff
@@ -1,535 +1,170 @@
...
 	case "", dstask.CMD_NEXT:
-		ts := dstask.LoadTasksFromDisk(dstask.NON_RESOLVED_STATUSES)
-		ts.Filter(context)
-		ts.Filter(cmdLine)
-		ts.FilterOutStatus(dstask.STATUS_TEMPLATE)
-		ts.SortByPriority()
-		context.PrintContextDescription()
-		ts.DisplayByNext(true)
-		ts.DisplayCriticalTaskWarning()
+		if err := dstask.CommandNext(repoPath, idsFilePath, stateFilePath, ctx, cmdLine); err != nil {
```
naggie, line 28: I think it feels a bit cumbersome to pass `repoPath, idsFilePath, stateFilePath` (and perhaps even the others) for every command. I saw your comment on why you added them -- I suppose the alternatives are:

1.  Use global variables like before
2. Keep this as part of the taskset (as discussed)

What do you think?

naggie: N.B. Otherwise I like the simplification affect this is having on dstask.go.

completions.go
```diff
@@ -0,0 +1,113 @@
+package dstask
```
naggie, line 1: It's nice to have this separated!

const.go
```diff
@@ -1,18 +1,14 @@
...
-import (
-	"os"
-	"os/user"
-	"path"
-	"strings"
-)
+import "os"
+
+func init() {
+	if os.Getenv("DSTASK_FAKE_PTY") == "1" {
```
naggie, line 6: I think just checking this is set to anything is best, as it might be confusing if set to some thing else (eg, `yes` is common)

const.go
```diff
@@ -113,13 +110,13 @@ var ALL_STATUSES = []string{
...
 // incomplete until all statuses are implemented
 var VALID_STATUS_TRANSITIONS = [][]string{
-	[]string{STATUS_PENDING, STATUS_ACTIVE},
-	[]string{STATUS_ACTIVE, STATUS_PAUSED},
-	[]string{STATUS_PAUSED, STATUS_ACTIVE},
-	[]string{STATUS_PENDING, STATUS_RESOLVED},
-	[]string{STATUS_PAUSED, STATUS_RESOLVED},
-	[]string{STATUS_ACTIVE, STATUS_RESOLVED},
-	[]string{STATUS_PENDING, STATUS_TEMPLATE},
+	{STATUS_PENDING, STATUS_ACTIVE},
```
naggie, line 113: Hm, didn't know this was allowable.

git.go
```diff
@@ -7,45 +7,45 @@ import (
...
-	args = append([]string{"-C", GIT_REPO}, args...)
+func RunGitCmd(repoPath string, args ...string) error {
+	args = append([]string{"-C", repoPath}, args...)
 	return RunCmd("git", args...)
 }
 
 // MustRunGitCmd delegates to RunGitCmd and exits the program on any error.
-func MustRunGitCmd(args ...string) {
-	err := RunGitCmd(args...)
+func MustRunGitCmd(repoPath string, args ...string) {
```
naggie, line 16: Quite a lot of variable passing again (I think I prefer the global variables / env, though not if it causes a testing issue)

const.go
dontlaugh: Me either, but `golint` and my vim plugin were shouting at me about it.

cmd/dstask.go
dontlaugh: There is another alternative that could reduce the number of parameters we take to the command functions in commands.go

We could consolidate these parameters into an application-level config struct.

```go
// config.go
type Config struct {
  Repo string `yaml:"repo"`
  IDsFile string `yaml:"ids_file"`
  StateFile string `yaml:"state_file"`
}
```

We then instantiate this on startup, using defaults as we do now so no one has to configure anything (unless they want to).

The config can be passed to the top level functions

```go
conf, _ := dstask.NewConfig(...)
if err := dstask.CommandNext(conf, ctx, cmdLine); if err != nil { /**/ }
```

For now, we could derive this struct from environment variables. But it lays the groundwork for a real config file, which I believe we will need eventually.

Of course, the lower level functions will look the same, as we will be passing values off this struct to the git commands, for instance.

naggie: That seems like a good compromise 👍

dontlaugh: I'll @ you again when the changes are up.

dontlaugh:
I'm implemented the `Config` struct in **config.go**.

I have not yet fixed the regression related to parsing the command line after the `--`.

dontlaugh:
This commit has a minimal test case that reproduces/fails for the regression naggie mentioned

https://github.com/naggie/dstask/pull/43/commits/f5947f07e64b5138238694e4dbabc7c279b13bd3

dontlaugh:
@naggie This is ready for re-review, but with a caveat. I've fixed the specific regression you mentioned with updates to cmdline.go, but have identified a different one (maybe). This is in a commented out test case here.
https://github.com/naggie/dstask/pull/43/files#diff-bdaeb5f4a2b8ebcaedc9316386f14ce7

My understanding is the `IDsExhausted` boolean is only used for completions, and that this is an edge case. But, should we open an issue to handle this case?

naggie:
`IDsExhausted` is used to stop parsing IDs -- IDs can be specified before or after a command (but not both) such that it's possible to add tasks with numbers in the description

dontlaugh:
Ah sorry I should clarify. From my grepping of the codebase, the only part of the code that uses `IDsExhausted` outside of the parsing function is the shell completions code. So it needs to be exported for that purpose, and tested, too.

naggie:
Oh sure I see what you mean. I guess it should be tested but we can handle that later. In the mean time I'm happy to merge, thanks for all the hard work and discussion!

PR #46, opened 2020-09-22
Build a statically linked binary for releases

arp242:
The README says:

> A single statically-linked binary

But it's not actually built as a statically linked binary:

    $ file ./dist/dstask-linux-amd64
    ./dist/dstask-linux-amd64: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), **dynamically linked**, interpreter /lib/ld-linux-x86-64.so.2 [..]

Also see `ldd` output.

This is probably because you're importing the net package (via
github.com/gofrs/uuid), as this contains a cgo-based DNS resolves which
gets compiled by default.

Just disabling cgo is the easiest way to ensure a static binary (or
error if your program really needs cgo). I wrote a bit more about this
on my website a while ago if you're interested:
https://www.arp242.net/static-go.html

### DIFF
--- a/do-release.sh
+++ b/do-release.sh
@@ -10,6 +10,7 @@ fi
 VERSION="$1"
 GIT_COMMIT=$(git rev-list -1 HEAD)
 BUILD_DATE=$(date)
+export CGO_ENABLED=0
 
 RELEASE_FILE=RELEASE.md
 


naggie:
Wow this is great, thanks very much @arp242 . I didn't know it was necessary to disable cgo. I'll read your article at lunchtime

naggie:
In fact, I'll do another release.

arp242:
Cheers; I saw someone mention it on Lobsters and it's an easy enough patch :-) Took me some time to figure all of this out last year as well, as the default behaviour can be kind of surprising (but it does make sense).

arp242:
As an aside, you can more than half the size by gzipping it:

    -rw-r--r-- 1 martin martin 3.5M Sep 22 23:59 dstask-linux-amd64
    -rw-r--r-- 1 martin martin 1.4M Sep 22 23:59 dstask-linux-amd64.gz

Granted, 2.1M isn't hugest of savings, but at my 100K/s download speeds it saves waiting a bit 😅 Should be easy enough to add it your release script.

naggie:
I specifically didn't want to add an extra step of decompression when installing and downloading, but thanks for the suggestion :+1:

PR #48, opened 2020-09-22
Update Makefile

tdslw:
This is a cold rewrite of the Makefile. Possibly not useful, because I might have not completly understood all the necessary steps or intentions. The new Makefile merges functionality from the old Makefile and do-release.sh. It introduces automatic variables offered by Make and tries place the prerequisites so that the dependencies are in the right place. Again, use this with care, if at all. Don't use this in a production environment, because it might break things accidentially. It was fun to write, and it would be great, if it is fun to use. If you get stuck with errors, but want to push through, drop me a line and I might see, if I could help.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,71 @@
-.PHONY: release clean unstall
-dist/dstask: clean
-	mkdir -p dstask
-	go build -mod=vendor -o dist/dstask cmd/dstask.go
+GIT_REPO   := github.com/naggie/dstask
 
-release:
-	./do-release.sh
+DIST_FILE  := dstask
+DIST_DIR   := dist
+SRC_FILE   := $(addsuffix .go, $(DIST_FILE))
+SRC_DIR    := cmd
+ARM5_FILE  := $(DIST_FILE)-linux-arm5
+AMD64_FILE := $(DIST_FILE)-linux-amd64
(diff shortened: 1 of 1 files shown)


dontlaugh:
This is a completely valid use of Make! But I (personally) am still unsure about this.

Right now, the only OS package I know of is for the AUR, and they do not make use of the makefile at all. We are lucky that dstask is a pure-go (no CGO) package right now, and a `go build` with appropriate flags is all that is required for packagers to deliver dstask to linux distros.

So, since the only known packager of dstask doesn't even need the Makefile, I would hesitate to introduce cross-compilation features before they are needed by distro packagers. The Makefile, at least for now, _seems_ to be a local development utility, and we are compiling natively for our respective architectures, and we don't need to set GOOS or GOARCH at all while doing that. Of course, Callan is the decider here.

naggie:
Thanks for the contribution @tdslw -- though I think @dontlaugh has a valid point about the current Makefile being redundant for general dstask building so I'd rather keep the simpler current system for now.

 I like the use of variables to build the target names.

tdslw:
The notifications for this somehow stuck in a filter for far to long, so I wasn't aware of your replies. As I said, it was also a interesting task on it's own. Keep up the good work. I will delete this branch at the end of the month.

Issue #44, opened 2020-09-14
task by index not working outside of context


naggie:
![image](https://user-images.githubusercontent.com/208440/93110068-bafa6000-f6ac-11ea-8ac5-ecbe179f6d02.png)

The default command, `next` should ignore the context if a task is referenced to be viewed by ID, much like the `edit` or `note` commands which work regardless of context if an ID is provided.

naggie:
@dontlaugh this may be awkward to fix with the [current filtering mechanism](https://github.com/naggie/dstask/blob/50474a1d4821d47eef7760491d7803f9032c76c5/taskset.go#L110) which sees a combined context and command line filters.

Perhaps the solution is not only ignore the context when an ID is provided, but the rest of the query too. We should probably also throw an error if an ID is specified together with command line filters.

dontlaugh:
I think you are right. I will try sketching out this solution:

* if IDs _and_ cmdLine tags or projects are passed, we error. This has the benefit of being catchable immediately after parsing, before a command is even selected
* of course, context tags will be allowed, since those could be set in the environment.
* if IDs are passed, we do not need to enter the filtering loop at all, we can take another code path and return early. This will ignore any other attributes in our query, as you describe

dontlaugh:
Actually, it would be cleaner (and reasonable?) to validate the options passed to NewTaskSet, as well. I will try to do that.

naggie:
thanks to @dontlaugh  -- fixed by https://github.com/naggie/dstask/pull/62

PR #57, opened 2020-10-07
Explicitly passed IDs take precedence over anything

dontlaugh:
We extract a function that sets the filtered attribute on tasks if the
task ID matches IDs explicitly passed at the command line. Add a unit
test for this logic.

Refs #44

### DIFF
--- a/cmd/dstask.go
+++ b/cmd/dstask.go
@@ -16,6 +16,11 @@ func main() {
 	ctx := state.Context
 	cmdLine := dstask.ParseCmdLine(os.Args[1:]...)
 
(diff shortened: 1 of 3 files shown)


dontlaugh:
@naggie I believe this should fix 44, or at least alleviate the situation.

Filtering relies on setting fields on an array of Task pointers, so I have created a private function to encapsulate this for IDs, and added a unit test.

We can do the same for filtering on other task attributes (tags, projects), in a follow up PR.

naggie:
See https://github.com/naggie/dstask/issues/59 -- I suggest this is merged into a new PR

PR #60, opened 2020-10-09
Add a suite of integration tests

dontlaugh:
These tests rely on a subset of our commands rendering an array of tasks to stdout as JSON
if stdout is not a TTY. Some tests are skipped because they cannot yet be used non-interactively
when stdout is not a TTY.

If we support other serialization formats, these tests will need to be updated.

Other changes:
* fix typo in Makefile
* vendor test helper libraries
* gitignore .vscode and .idea editor/IDE metadata


integration/log_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"testing"
+
+	"github.com/naggie/dstask"
+	"gotest.tools/assert"
+)
+
+func TestLog(t *testing.T) {
+	t.Skip("skipping because show-resolved cannot be run non-interactively yet")
```
dontlaugh, line 11: Skipping this for now. We could remove skip once we can run show-resolved w/o a TTY.

integration/modify_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"testing"
+
+	"github.com/naggie/dstask"
+	"gotest.tools/assert"
+)
+
+func TestModify(t *testing.T) {
+	t.Skip("modify test TODO")
```
dontlaugh, line 11: @naggie skipping this one also. What is the expected behavior for modify with respect to IDs and context?

Is it:

* 1 or more IDs always takes precedence, no matter what context is set
* if a context is set (and no IDs provided), a modify adds a tag or project to every task in the context?

What is an example of a modify command outside context? e.g. involving the `--`

integration/show_open_test.go
```diff
@@ -0,0 +1,3 @@
+package integration
+
+// TODO: what is a good integration test for "show-open"?
```
dontlaugh, line 3: @naggie Can you recommend a strategy for these tests?

integration/show_templates_test.go
```diff
@@ -0,0 +1,3 @@
+package integration
+
+// TODO
```
dontlaugh, line 3: I haven't actually used the template feature yet. Do i need to add a task first, then turn it into a template? Or, can I do this in one go?

integration/log_test.go
naggie: What's the reason we can't? Just not implemented, I assume.

integration/modify_test.go
naggie: Yes, correct. I suppose naturally the `--` would work as if a context was not set -- i.e. if there are no IDs, set to all non-resolved tasks.

integration/show_open_test.go
naggie: I suppose just checking that the expected number of tasks are presented for the given context. I suppose with JSON output, as there's no terminal height limitation the `next` and `show-open` are effectively the same.

integration/show_templates_test.go
naggie: I think the template command creates a new template from scratch, so you can do it in one go. That means [this commen](https://github.com/naggie/dstask/blob/75615db7f53b16e51a4bd08bce99d9cf7815ef88/commands.go#L570)t is incorrect.

integration/log_test.go
dontlaugh: Correct, merely that it isn't implemented yet

dontlaugh:
Thanks for taking a look. I will try to address the TODOs today. I'll try to determine how to get the show-resolved command printing JSON as well.

naggie:
No prob. Great, thanks! :)

integration/modify_test.go
dontlaugh: I have added a test for modifying by IDs, but we will have to implement context-based modify in a later PR. A prompt is issued if an ID is not set:

```
--- FAIL: TestModify (0.06s)
    modify_test.go:31: stdout: 
    modify_test.go:31: stderr: No IDs specified. Apply to all 1 tasks in current ctx? [y/n] panic: EOF
```

dontlaugh:
@naggie Okay, I believe I've addressed everything discussed here, with the exception of a test around `modify` when IDs are *not* provided. We will need to skip the interactive prompt in this case.

naggie (approved):

naggie:
Seems good, I'll run it for a bit then merge. Thanks!

PR #61, opened 2020-10-16
Allow non-interactive modify against entire context

dontlaugh:
We change the behavior of the modify command to only prompt if stdout is
a TTY. This lets us add an integration test for modify that tests
modifying all tasks within an active context.


naggie:
Good idea, thanks!

Issue #59, opened 2020-10-08
aggregate issues for filtering refactor


naggie:
The filtering refactor here https://github.com/naggie/dstask/pull/53 was merged but later reverted due to significant bugs. I think we're best working on a separate branch/PR until the bugs are fixed and tested

- [ ] task addressing by ID is broken
- [ ] the modify command ignores the ID passed to it and applies the change to every given task. (probably caused by the above)
- [ ] search by keyword is broken

I think given this is a refactor of filtering it would make sense to fix these in this branch too:

- [ ] https://github.com/naggie/dstask/issues/44 task by index not working outside of context
- [ ] https://github.com/naggie/dstask/pull/57 Explicitly passed IDs take precedence over anything (which doesn't yet account for the `modify` command, I think).

We should have unit tests for the above too.

@dontlaugh please could you create a new PR for #56 then we can work on those issues. Thanks

dontlaugh:
Sounds good to me. :+1: I will make sure all these things are taken into account.

dontlaugh:
How are we to determine if a search should be added into our filtering logic?

Is it: if _any_ text is present in the `Text` field of a parsed CmdLine AND if the Cmd is a "read" command (such as "next"), then that text is to be used as a filter? Meaning, any tasks left over after context/tags/projects are filtered, will be further filtered by a search for the `Text` string passed?

naggie:
Yes exactly, text is always part of the filtering logic. It shouldn't be present when using the modify command though.

dontlaugh:
An end-to-end integration test would help prevent these regressions, and might be easier to implement and maintain in the near term. 

Would you mind if I cherry-picked the JSON rendering from #58 and introduced it for the commands that render complete tasks?

* show-active
* show-paused
* show-open
* next (also the default command)
* show-templates
* show-resolved
* show-unorganised

I definitely want unit tests, as well. But unit testing itself requires refactoring. In the meantime, it might be convenient to be able to have a test script like this to watch our backs.

```sh
set -e
dstask add +foo task one
dstask add +baz task two
# a non-zero exit code from diff will signal an error to us.
diff <(dstask) ./integration/testfiles/expected.json
```

As long as our assignment of IDs remains deterministic, this could also work for `modify` as well. 

[...]

PR #62, opened 2020-10-18
Bring sorting and filtering into NewTaskSet

dontlaugh:
This is a rework of #53 but this time guided by the integration tests and integrating
some bugfixes along the way.

The bugs were related to parsing and applying (or not applying) filters inside and
outside of a context. There is probably more testing we can do to refine our exact
specifications, and I am hoping that our `TaskSetOpts` can handle that.

The bugs are aggregated in #59 so we should make sure those things are covered
before merging. See also #44

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -275,6 +275,7 @@ func CommandNext(conf Config, ctx, cmdLine CmdLine) error {
 		WithTags(cmdLine.Tags...),
 		WithoutTags(ctx.AntiTags...),
 		WithoutTags(cmdLine.AntiTags...),
+		WithText(cmdLine.Text),
 	)
 	if err != nil {
 		return err
--- a/commands.go
+++ b/commands.go
@@ -78,11 +78,11 @@ func CommandContext(conf Config, state State, ctx, cmdLine CmdLine) error {
 		fmt.Printf("Current context: %s\n", ctx)
 	} else if os.Args[2] == "none" {
 		if err := state.SetContext(CmdLine{}); err != nil {
-			ExitFail(err.Error())
+			return err
 		}
 	} else {
 		if err := state.SetContext(cmdLine); err != nil {
-			ExitFail(err.Error())
+			return err
 		}
 	}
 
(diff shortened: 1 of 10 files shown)


dontlaugh:
marking this as WIP while I work through the bugs

commands.go
```diff
@@ -246,17 +262,24 @@ func CommandModify(conf Config, ctx, cmdLine CmdLine) error {
...
 		WithStatuses(NON_RESOLVED_STATUSES...),
+		WithIDs(cmdLine.IDs...),
+		WithProjects(ctx.Project, cmdLine.Project),
+		WithoutProjects(ctx.AntiProjects...),
+		WithoutProjects(cmdLine.AntiProjects...),
+		WithTags(ctx.Tags...),
+		WithTags(cmdLine.Tags...),
+		WithoutTags(ctx.AntiTags...),
+		WithoutTags(cmdLine.AntiTags...),
+		WithText(cmdLine.Text),
```
dontlaugh, line 278: We pass command line Text into the constructor as well, since it can technically also act as a filter in the context of the `next` command.

integration/next_search_test.go
```diff
@@ -13,12 +13,13 @@ func TestNextSearchWord(t *testing.T) {
 
 	program := testCmd(repo)
 
-	output, exiterr, success := program("add", "one")
+	output, exiterr, success := program("add", "one", "/", "alpha")
```
dontlaugh, line 16: add a string to the Note field here

smoketest.sh
```diff
@@ -57,7 +56,12 @@ git -C $UPSTREAM_BARE_REPO init --bare
 ./dstask next
 ./dstask 1 done
 ./dstask show-resolved
+
+# TODO we set FAKE_PTY because we do not have a non-tty
+# rendering scheme for show-projects
+export DSTASK_FAKE_PTY=1
 ./dstask show-projects
+unset DSTASK_FAKE_PTY
```
dontlaugh, line 64: It seems show-projects is the only command that still requires the DSTASK_FAKE_PTY variable. We might open an issue to provide a scriptable output for this command, as we've done with others. JSON is what we've done for the others, and that might work here, as well.

dontlaugh: https://github.com/naggie/dstask/issues/63

integration/next_search_test.go
naggie: So searching notes is a separate thing? Right now searching with dstask searches the description and note implicitly. I think this works well given the fairly limited scope of searching non-resolved tasks by default. What do you think?

dontlaugh: Here we simply add a note to task one, but below we do check for
* searching for something that isn't there (no results)
* finding string "alpha" in a note
* finding task "two" via its summary

Is there another test we should add? The strings are admittedly really simple.

naggie: Sorry I misread, late at night. Seems fine as is.

naggie:
Thanks @dontlaugh ! -- working great and solves the issues.

PR #65, opened 2020-10-24
Bugfix: combining tags should use AND logic

dontlaugh:
For example: if we pass a filter or set a context of "+one +two", we
should include tasks that have tags "+one", "+two", or both.

### DIFF
--- a/integration/next_context_test.go
+++ b/integration/next_context_test.go
@@ -42,3 +42,36 @@ func TestSettingTagContext(t *testing.T) {
 	tasks = unmarshalTaskArray(t, output)
 	assert.Equal(t, tasks[0].Summary, "two", "setting -one as a context")
 }
+
+func TestSettingTagAndProjectContext(t *testing.T) {
+	repo, cleanup := makeDstaskRepo(t)
+	defer cleanup()
+
+	program := testCmd(repo)
+
+	output, exiterr, success := program("add", "+one", "+alpha", "one")
+	assertProgramResult(t, output, exiterr, success)
+
+	output, exiterr, success = program("add", "project:beta", "+two", "two")
+	assertProgramResult(t, output, exiterr, success)
+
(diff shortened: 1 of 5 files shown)


util.go
```diff
@@ -125,6 +125,22 @@ func StrSliceContains(haystack []string, needle string) bool {
 	return false
 }
 
+func StrSliceContainsAll(subset, superset []string) bool {
```
dontlaugh, line 128: For tag filtering logic, I am assuming we need to compare two slices. In our case a `WithTags` filter is the subset, and the existing list of `Tags` is the superset.

Everything passed as `WithTags` needs to be contained within existing `Tags`, otherwise we filter.

integration/next_context_test.go
```diff
@@ -42,3 +42,36 @@ func TestSettingTagContext(t *testing.T) {
 	tasks = unmarshalTaskArray(t, output)
 	assert.Equal(t, tasks[0].Summary, "two", "setting -one as a context")
 }
+
+func TestSettingTagAndProjectContext(t *testing.T) {
```
dontlaugh, line 46: Strictly speaking, this test is not related to the bugfix, but we were missing a minimal example related to setting `project:` attributes.

taskset.go
```diff
@@ -136,28 +138,29 @@ func NewTaskSet(repoPath, idsFilePath, stateFilePath string, opts ...TaskSetOpt)
...
 
+		// if a WithoutProjects option is passed, our task's project cannot
+		// be contained within it
 		for _, antiProject := range tso.withoutProjects {
 			if antiProject == task.Project {
 				task.filtered = true
 			}
 		}
 
-		for _, tag := range tso.withTags {
```
dontlaugh, line 145: This diff here is the bufix (I believe).

naggie:
wow good catch, I hadn't tested that recently. Thanks. I'd better make a release.

PR #66, opened 2020-10-31
Allow setting context with DSTASK_CONTEXT

dontlaugh:
If this env var is set to a valid context string, it takes precedence
over context stored in local state. The "--" ignore context token will
still be respected, however.

Refs #20

### DIFF
--- a/integration/next_context_test.go
+++ b/integration/next_context_test.go
@@ -1,6 +1,7 @@
 package integration
 
 import (
+	"os"
 	"testing"
 
 	"github.com/naggie/dstask"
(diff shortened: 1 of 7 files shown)


integration/next_context_test.go
```diff
@@ -75,3 +76,33 @@ func TestSettingTagAndProjectContext(t *testing.T) {
...
+
+	output, exiterr, success = program("add", "project:beta", "+two", "two")
+	assertProgramResult(t, output, exiterr, success)
+
+	output, exiterr, success = program("context", "project:beta")
+	assertProgramResult(t, output, exiterr, success)
+
+	// override context with an env var
+	unsetEnv := setEnv("DSTASK_CONTEXT", "+one +alpha")
+	//unsetEnv := setEnv("DSTASK_CONTEXT", "+one")
```
dontlaugh, line 97: I can add more unit tests to confirm this, but setting a space-containing string like "+one +alpha" appears to _parse_ correctly, but still fails our integration test.

Setting a string without spaces (just a single tag in this case) works  fine. I kind of expected both to work.

Maybe there's something about env vars in unix I don't understand.

dontlaugh: Ah this might be because the full string "+one +alpha" is passed as a single argument to `ParseCmdLine`. At this point, our parsing logic assumes that the command line has been tokenized, and interprets this as a single tag.

dontlaugh: That was the problem. [This standard library function](https://golang.org/pkg/strings/#Fields) did the trick.

dontlaugh:
@naggie I opted for a shorter env var name. Let me know if there's another you'd prefer.

naggie:
Just a few thoughts:

* Could mention the use cases in an advanced section of the readme (eg, direnv for automatic setting of `project:` inside a git repository)
* I think we should warn/notify the user (yellow) when the env var is being used (I;m just a bit wary of this causing a bit of confusion or accidental tags etc)
* Attempting to set a context when the env var is set should result in an error

I think I might set my bash/zsh prompt to display the current dstask context like it shows the git branch.

dontlaugh:
How about this?

image
This is one way to do it. The message can be shimmed in here, otherwise it will be empty string if DSTASK_CONTEXT is not set https://github.com/naggie/dstask/pull/66/commits/5b9239c315455639fd619e4de0b09a563addaa87

dontlaugh:
Here is an implementation of the error case you described

image

dontlaugh:
For the documentation bit, I opted for a whole section just about Context.

naggie (approved):

naggie:
Thanks @dontlaugh looks top notch.

Issue #67, opened 2020-11-01
Better message when the table is truncated. 


xm31:
When the size of the terminal doesn't permit to display all the tasks, It will be nice to see also the number of displayed tasks.
For now,  only the total number of tasks and the number of rows are shown.

naggie:
Merged, thanks.

xm31:
Great! And thank you for doing this. I was looking for a replacement for task warrior for a while. Now I found it... :)

naggie:
You're welcome, it's nice to have good feedback. Out of interest, what made you change from taskwarrior?

xm31:
The sync part did not go well for me. I tried [taskd](https://wiki.archlinux.org/index.php/Taskd) (not exactly the easiest way to set it up) and hooked git commands (fine until there are conflicts which is very common with recursive daily tasks).

Also, I always felt that it would be nice to be able to add more content to a task.

naggie:
Ah, that reflects my own experience and is why I wrote dstask :+1:

PR #71, opened 2020-11-04
display similar message with taskwarrior when the table is truncated.…

xm31:
… #67

### DIFF
--- a/display.go
+++ b/display.go
@@ -117,7 +117,7 @@ func (ts *TaskSet) renderTable(truncate bool) error {
 		table.Render()
 
 		if truncate && maxTasks < total {
-			fmt.Printf("\n%v tasks, truncated to %v lines.\n", total, maxTasks)
+            fmt.Printf("\n%v/%v tasks shown.\n", maxTasks, total)
 		} else {
 			fmt.Printf("\n%v tasks.\n", total)
 		}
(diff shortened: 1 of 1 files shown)


display.go
```diff
@@ -117,7 +117,7 @@ func (ts *TaskSet) renderTable(truncate bool) error {
 		table.Render()
 
 		if truncate && maxTasks < total {
-			fmt.Printf("\n%v tasks, truncated to %v lines.\n", total, h)
+			fmt.Printf("\n%v tasks, truncated to %v lines.\n", total, maxTasks)
```
naggie, line 120: the message should probably also be changed to something that says tasks instead of lines, maybe `fmt.Printf("\n%v tasks, showing to %v.\n", total, maxTasks)`

xm31: What do you think about  `fmt.Printf("\n%v tasks, %v of them displayed.\n", total, maxTasks)` ?

naggie: `of them` is probably superfluous, otherwise sounds good!

xm31: Yeah. You're right. It is superfluous. :)
The last change in the pull request is with 
`fmt.Printf("\n%v/%v tasks shown.\n", maxTasks, total)`

naggie: Thanks, will merge after work.

Issue #69, opened 2020-11-02
show-resolved shows some tasks with invalid date


naggie:
Since the filtering updates, `show-resolved` is very broken.

1. It shows an invalid date `> Week 1, starting Mon 1 Jan 0001`.
2. It ignores the current context.

dontlaugh:
If you can tell, does the data on disk look correct?

I'll take a look at this after work.

naggie:
Using dstask v0.18 things work correctly, so the data is fine. Thanks!

dontlaugh:
This may be a problem with CommandDone

Should we set the resolved timestamp here?
https://github.com/naggie/dstask/blob/master/commands.go#L104-L110

CommandLog does it correctly
https://github.com/naggie/dstask/blob/master/commands.go#L201

dontlaugh:
A remedy for existing data may be using the unix filesystem mod time as a fall back for a bad resolved date. This mod time should be

* preserved/transferred by git
* _usually_ the same as when the task was resolved, since that is the likely cause of a file being created under the resolved/ folder

https://golang.org/pkg/os/#FileInfo from Stat() can give us this.

naggie:
True -- though I'd rather not rely on the modification date in a prod code path as it's normally wrong. Unless we use it to set the date with a migration. Alternatively we could add a python script to correct the data in `etc/` -- or just leave it, the fall out is minimal.

dontlaugh:
@naggie I was thinking a one-time migration, myself. Migrations are something we might need to come to terms with.

Leave this issue open. I might see what a migration looks like while I check on the context portion of this bug.

Sorry for the grief.

naggie:
True, I guess we can say to run a command as part of the release notes.

> Sorry for the grief.

No problem, bugs are part software dev and you're adding regression tests 👍

naggie:
Perhaps a hidden `dstask migrate` command could be used, and only documented in release notes.

dontlaugh:
That would be a good way to invoke it. We may consider a one-time prompt, for users who miss the notes.

I think the fix I proposed earlier won't be hard to implement, and I won't delay that by insisting that we design a full-blown migration framework. I think we'll want one at some point, but for now a `dstask migrate` command looks good.

If someone experiences this bug, at this stage they will probably come looking here in the repository.

dontlaugh:
So a nice thing about using git: we can run migrations for bug fixes, but not commit. We could leave the directory with changes and ask the user to commit.

[...]

PR #72, opened 2020-11-06
Migrations directory, new migration 001

dontlaugh:
This is a bugfix for #69


dontlaugh:
Was kind of hoping for a sub-package, but cyclic imports bit me. We definitely need the `dstask.Task` type to deserialize our files into structured data.

dontlaugh:
Some testing shows no more resolved tasks from ancient times https://gist.github.com/dontlaugh/f8da6fecb0159d8555964136d13d34db#file-log-txt-L37-L56

dontlaugh:
@naggie ready for review. `dstask migrate` should make changes to tasks with the zero value under the resolved directory, and commit the result with a message about the migration.

dontlaugh:
We have decided to close this. If anyone comes across the bug in #69 the fix is straightforward, and I can help with writing a script to do it. But we don't need to check in a migration framework written in Go, yet.

PR #97, opened 2020-12-01
Filtering changes

naggie:
Some bits from https://github.com/naggie/dstask/issues/90

- [x] `cmdLine` is really a query, so should be changed to `Query`. It just so happens that one place it's derived from is the command line (another is context)
- [x] `TaskSetOpts` is basically the same thing as CmdLine/Filter but with a sort order. Factor out.
- [x] Filters in taskset constructor -- change to use in Filter/Sort methods like before. Usage is inconsistent at the moment. Combining context with cli Query worked well.
- [x] Better rule checking for given query (eg, `done` should be given IDs and optional note only)
- [x] edit command -- reattempt mechanism

---

I'll stop here -- I plan on tweaking a few things like the `template` command names/behaviour and the `start` behaviour too. Will discuss in slack.

Unit tests + QA tests were invaluable. Thanks guys.


commands.go
```diff
@@ -12,36 +12,36 @@ import (
 )
 
 // mergedTaskSetOpts returns a TaskSetOpt that applies the various filters
-// that should be exerted by the ctx and cmdLine
-func mergedTaskSetOpts(ctx, cmdLine CmdLine) TaskSetOpt {
+// that should be exerted by the ctx and query
+func mergedTaskSetOpts(ctx, query Query) TaskSetOpt {
```
Dieterbe, line 16: alternatively we could also make a Query.Merge(Query) method that returns a new merged query.

dontlaugh: I think an exported method w/ a unit test is preferable, since we probably want to decide on an official strategy for merging these objects.

naggie: > alternatively we could also make a Query.Merge(Query) method that returns a new merged query.

You mean like this? https://github.com/naggie/dstask/blob/d687a0f9aec634a808cfa0df4aecb06ab4f57732/task.go#L66

That's how it was before the refactor; I am restoring it. 👍

dontlaugh (commented):
Looks good to me. I would just clean up the doc strings.

query.go
```diff
@@ -152,3 +163,39 @@ func ParseCmdLine(args ...string) CmdLine {
...
+
+	for _, tag := range q2.AntiTags {
+		if !StrSliceContains(q.AntiTags, tag) {
+			q.AntiTags = append(q.AntiTags, tag)
+		}
+	}
+
+	if q2.Project != "" {
+		if q.Project != "" && q.Project != q2.Project {
+			ExitFail("Could not apply q2, project conflict")
```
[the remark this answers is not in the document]
naggie: Good point, I missed that when copying it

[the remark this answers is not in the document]
naggie: I'll make an issue for that and deal with it in another PR 👍

naggie:
Thanks for the review @dontlaugh

Issue #100, opened 2020-12-18
Task Urgency


cgardner:
Pulling this out of https://github.com/naggie/dstask/issues/90#issuecomment-731606087

Now that I've been using dstask for a while, I'm finding that the urgency and sorting by urgency that taskwarrior did was very helpful for my own prioritization.  It allowed me to work from the top of the list and not have to think about what the most important thing was.

> ## Priority/urgency
> Taskwarrior calculates urgency based on priority and heuristics from weightings gathered from projects/tags. We don't want to make something too complicated or overconfigurable. May change later, for instance considering tasks as part of projects more urgent.

PR #101, opened 2020-12-31
Urgency

cgardner:
Addressing #100 

- [x] Add Urgency to the Task object
- [x] Only set urgency for open tasks
- [x] Add the ability to sort by urgency, grouped by priority

### DIFF
--- a/doc/urgency.md
+++ b/doc/urgency.md
@@ -1,0 +1,12 @@
+# Urgency
+
+Urgency is pretty primitive at the moment. It should only affect tasks that are
+not resolved.
+
+It takes the Task's priority as a Fibonacci (Low = 1, Critical = 5) and an applies a multiplier (5).
+
+If the status is active, it adds 5.
+
+If a project or tag is applied, add 3 for each.
+
+It also multiplies the age of the task by 0.05 and adds that to the urgency.
--- a/integration/next_id_index_test.go
+++ b/integration/next_id_index_test.go
@@ -74,6 +74,12 @@ func TestNextWithurgency(t *testing.T) {
 	output, exiterr, success = program("next")
 	assertProgramResult(t, output, exiterr, success)
 	tasks := unmarshalTaskArray(t, output)
+
+	// This order is expected.
+	// - three is a P0 task
+	// - two has a project and a tag
+	// - four just has a project
+	// - one doesn't have a project or tag
 	assert.Equal(t, "three", tasks[0].Summary)
 	assert.Equal(t, "two", tasks[1].Summary)
 	assert.Equal(t, "four", tasks[2].Summary)
--- a/integration/next_id_index_test.go
+++ b/integration/next_id_index_test.go
@@ -52,3 +52,30 @@ func TestNextByIDIndexOutsideContext(t *testing.T) {
 	tasks = unmarshalTaskArray(t, output)
 	assert.Equal(t, 1, tasks[0].ID, "1 is the only ID in our current context")
 }
+
+func TestNextWithurgency(t *testing.T) {
+	repo, cleanup := makeDstaskRepo(t)
+	defer cleanup()
+
+	program := testCmd(repo)
+
+	output, exiterr, success := program("add", "one")
+	assertProgramResult(t, output, exiterr, success)
+
+	output, exiterr, success = program("add", "two", "+two", "project:two")
+	assertProgramResult(t, output, exiterr, success)
+
+	output, exiterr, success = program("add", "three", "P0")
+	assertProgramResult(t, output, exiterr, success)
+
+	output, exiterr, success = program("add", "four", "project:four")
+	assertProgramResult(t, output, exiterr, success)
+
+	output, exiterr, success = program("next")
+	assertProgramResult(t, output, exiterr, success)
+	tasks := unmarshalTaskArray(t, output)
+	assert.Equal(t, "three", tasks[0].Summary)
+	assert.Equal(t, "two", tasks[1].Summary)
+	assert.Equal(t, "four", tasks[2].Summary)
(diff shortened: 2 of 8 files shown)


kingo55:
This looks useful.

Looks like we could use task notes for additional context based on Task Warrior's formula [1]. And eventually task dependency/due dates.

1: https://taskwarrior.org/docs/urgency.html

dontlaugh:
Can you add an integration test for this? Maybe a description of how urgency is computed?

task.go
```diff
@@ -301,6 +305,52 @@ func (task *Task) Modify(cmdLine CmdLine) {
...
+func (task *Task) IsResolved() bool {
+	for _, nonResolvedStatus := range NON_RESOLVED_STATUSES {
+		if task.Status == nonResolvedStatus {
+			return false
+		}
+	}
+	return true
+}
+
+func generateUrgency(task Task) int {
```
dontlaugh, line 317: We might prefer `computeUrgency` or `determineUrgency`, to imply that it's deterministic. This also looks like it should be unit tested, since it's a bit complex

cgardner: I changed the name and added some unit tests.  I also added a make task to generate code coverage so I could make sure I had all cases covered.

integration/next_id_index_test.go
```diff
@@ -52,3 +52,30 @@ func TestNextByIDIndexOutsideContext(t *testing.T) {
...
+
+	output, exiterr, success = program("add", "three", "P0")
+	assertProgramResult(t, output, exiterr, success)
+
+	output, exiterr, success = program("add", "four", "project:four")
+	assertProgramResult(t, output, exiterr, success)
+
+	output, exiterr, success = program("next")
+	assertProgramResult(t, output, exiterr, success)
+	tasks := unmarshalTaskArray(t, output)
```
dontlaugh, line 76: A comment here would be good:
"We expect this sort order because X, Y, Z..."

naggie:
so @cgardner  -- this PR seems well written. My only concern is the calculation seems to be a bit arbitrary vs the fairly basic sorting we already have. I'd like to have some way of determining the "correct" factors.

Also glad to see coverage in there, it will be good to put that in master.

cgardner:
It's absolutely arbitrary. I went with values that felt right to me.  I figured we could discuss and customize that bit during the review process.

I will pull out the coverage work to a separate PR so it can move independently.

doc/urgency.md
```diff
@@ -0,0 +1,12 @@
+# Urgency
+
+Urgency is pretty primitive at the moment. It should only affect tasks that are
+not resolved.
+
+It takes the Task's priority as a Fibonacci (Low = 1, Critical = 5) and an applies a multiplier (5).
```
Dieterbe, line 6: this is very difficult for a layman (including myself) to understand.

naggie: Agreed, and the reasoning behind it doesn't come naturally. 

I suggest an approach of identifying then addressing individual problems that the current sorting method has; note that if the basic method is good 95% it's probably a good solution

doc/urgency.md
```diff
@@ -0,0 +1,12 @@
...
+
+If the status is active, it adds 5.
+
+If a project or tag is applied, add 3 for each.
+
+It also multiplies the age of the task by 0.05 and adds that to the urgency.
```
Dieterbe, line 12: should it not take into account the due date and the task priority?

cgardner: Priority is taken into account.  After adding urgency, it would sort by priority > urgency > created.

I didn't do anything with due date because I understood it wasn't implemented, but it shouldn't be too hard to add that.

doc/urgency.md
cgardner: This was a complete guess to start out.  I left the implementation mostly as a way to start the conversation about what the weights should be for everything.

cgardner:
I updated the output to show a table of why the urgency is what it is.  Hopefully this will make the reason more transparent.  I can add the full calculation to the table if that would be more helpful.

image

cgardner:
Nothing specific, just more fine grained sorting.  I could go either way, but learned a lot doing this. :)

naggie:
Ok then -- thanks for the investigation nonetheless!

ecocode:
I would love this urgency feature, but somewhat it should be user-configurable.  Perhaps like taskwarrior does.
ie. With taskwarrior tasks with the "next" tag will show high up in the list.  This is very useful to figure which task to work on next, prject indepently.
When using the user defined P0-4 priorities, tasks often all end up in P0, because you tend to put a task higher in the list by increasing the priority.  You can degrade priority of current P0 tasks to make the "next" tagged task get above the others, but that looks like a cumbersome job.
Of course you can perhaps script this.  But then again there are other stuff you want to show higher in the list.  Like ie. I have 3 projects which are urgent because of the due date coming near.  With taskwarrior you can then just add a specific factor to those projects, even finetune priorities between them.

Definitely a wishfull feature !

ecocode:
@cgardner Can you consider reopening this enhancement request based on my previous comment ?

niderhoff:
> I'm not yet convinced that it needs to be more complicated than the very straightforward priority/age thing where we effectively rely on the user to set an urgency via priority

I think it's super helpful if you can add labels or projects and that will automatically give an urgency. The label<>urgency combo in taskwarrior is really powerful. Instead of thinking in "priorities" (which is very low level) I can think in more high level attributes to organize my task and then I can just have the program figure out the ranking of my tasks automatically.

Also it gives you a discrete scale from 0 to infinity which much more room than just using 1-4 scale.

Think about it this way: I want to define a priority of projects and have the program figure out the order. I don't want to define the order myself.

Issue #116, opened 2021-01-23
Make git branch name configurable


kevindkeogh:
Hi all,

Great tool.  I typically use `main` as my main git branch name, and noticed that `master` is hardcoded in the `dstask sync` command. Was wondering if it was possible to make the branch name configurable?

https://github.com/naggie/dstask/blob/05f55e1c0c50dc7937ddf9e778be625bbd551609/git.go#L95

Thank you!

dontlaugh:
@kevindkeogh If you are willing to build a custom branch, see my fork in #118 and see if it fixes your issues.

I am having a heck of a time getting the tests to pass, though. Some kind of Go modules issue.

kevindkeogh:
This worked perfectly for me, thank you!

If its any help, I ran
```
git clone https://github.com/gloriousfutureio/dstask.git
cd dstask
git checkout rely_on_default_branch 
go mod vendor
make
./dist/dstask sync
```

```
$ go version
go version go1.15.7 linux/amd64
```

dontlaugh:
@kevindkeogh  I spoke to naggie and we're taking a different approach.

We will ask users to set a `DSTASK_BRANCH` environment variable to change the git branch. The default will be master, for backwards compatibility.

This is just a heads up. See the associated PR if you're interested.

adamsoderstrom:
@naggie 

Hm... maybe! According to [Git's "Remote Branches" documentation](https://git-scm.com/book/en/v2/Git-Branching-Remote-Branches#_tracking_branches), it appears as you could retrieve the remote branches via `git branch -vv`.

If one would change branch, then one would probably expect the tasks to be pushed on that branch..? 🤔

dontlaugh:
It has been a while since I've thought about this issue, specifically. However, I've spent quite a few hours writing automation around git for various day jobs, and my inuition is that there isn't a _robust_ way to do this without adding extra config. Git is fundamentally distributed, and I think there will always be some failure modes if we are to rely on `git config` itself to carry this little bit of state around (the branch name).

naggie:
I suppose we could list origin branches, and if the list contains `main` but not `master` pick `main` -- and nice versa.

Or, require an upstream tracking branch to be configured for the current branch, and just push/pull from that.

naggie:
I suppose we could list origin branches, and if the list contains `main` but not `master` pick `main` -- and nice versa.

Or, require an upstream tracking branch to be configured for the current branch, and push/pull from that implicitly

[...]

PR #118, opened 2021-01-26
Remove branch name from push

dontlaugh:
It is possible to specify an alternative default initial branch for git

Refs #116


dontlaugh:
I _think_ this might be an easy fix for #116, but without adding any configuration options. If users have an alternative **init.defaultBranch** variable set, it will require zero intervention. If they don't, it will mean just checking out a different branch and setting a remote tracking branch.

PR #129, opened 2021-02-23
#128 inline template notes

### LINKED ISSUE #128: Using "/" notation to add notes to a new task with a template doesn't work.
I found out that creating a task with a template and notes doesn't add the notes passed in from the command line.

```bash
dstask add "Test Task" template:123 / This should be a note
```

I could see this as acceptable if the template has default note, but it's for all templates.

cgardner:
Closes #128

### DIFF
--- a/task.go
+++ b/task.go
@@ -261,12 +261,10 @@ func (t *Task) Modify(query Query) {
 		t.Priority = query.Priority
 	}
 
-	if query.Note != "" {
-		if t.Notes != "" {
-			t.Notes += "\n"
-		}
-		t.Notes += query.Note
+	if t.Notes != "" {
+		t.Notes += "\n"
 	}
+	t.Notes += query.Note
 }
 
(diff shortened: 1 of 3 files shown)


task.go
```diff
@@ -260,6 +260,13 @@ func (t *Task) Modify(query Query) {
 	if query.Priority != "" {
 		t.Priority = query.Priority
 	}
+
+	if query.Note != "" {
```
naggie, line 264: I think the if statements can be combined, as if query.Note is `""` when appending it to `t.Notes` will have no effect so it doesn't have to be conditional; the only thing that needs to be is the addition of `"\n"`.

naggie: Other than that looks good to go. Well tested too. Thanks!

dontlaugh:
@cgardner Anything else on this?

cgardner:
I don't think so

naggie:
Apparently I forgot about this, sorry!

PR #139, opened 2021-05-07
fix link to dstask-import tw instructions

axelsimon:
No need to add a `/doc/` dir, `dstask-import.md` is in the same dir as `taskwarrior-migration.md`.


naggie:
Oops -- thanks @axelsimon

PR #149, opened 2021-08-05
Installation: fix link to releases

axelsimon:
also, capitalise an instance of Taskwarrior and add a link to https://www.passwordstore.org/, because why not


naggie:
Thanks @axelsimon

PR #150, opened 2021-08-11
Restore original ordering 

naggie:
dstask originally had the eldest tasks shown first, at the top. I think this makes sense as the older tasks probably should be addressed first if no other prioritisation information is known. 

This PR restores the original behaviour.


Dieterbe (approved):
agreed

PR #152, opened 2021-09-18
Require --ff on pull

dontlaugh:
A globally-set --ff-only means a git pull will refuse to fall back to
merge commit if a fast forward cannot be set. This causes sync errors
more frequently.

Let's prefer setting --ff explicitly, making the choice on behalf of the
user to prefer merge commits to failures. The only downside is the
non-linear commit history, but this seems minor, given our use case.

See `git pull --help`


naggie (approved):
Thanks, a useful change

PR #160, opened 2021-10-23
Encode Task ID and Summary in EDITOR tempfile

### LINKED ISSUE #117: Encode (part of) task ID and Summary in editor tempfile name
I often have an editor buffer open for a long period of time, or I leave the terminal and come back and it's been open for a while.

Most editors display the name of the open file. Currently, if I do `dstask edit 1`, the name is mostly random:

```
/tmp/dstask.12345231.tmp
```

I think encoding a fixed-width portion of the ID and Summary could be useful. So the above, for ID 1 and a Summary "Email Bobby about birthday party" could be edited with a temp file like

```
/tmp/dstask.1_email_bobby_ab.12345231.tmp
```

dontlaugh:
Introduce a new utility function (and unit test) for generated truncated
filenames from task ID and summary.

Go strings are UTF-8, and there may be pathological inputs that we need
to sanitize. Add these to the unit test and handle them as we find them.

Fixes #117

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -146,7 +146,7 @@ func CommandEdit(conf Config, ctx, query Query) error {
 		}
 
 		for {
-			edited := MustEditBytes(data, "yml")
+			edited := MustEditBytes(data, MakeTempFilename(task.ID, task.Summary, "yml"))
 			err = yaml.Unmarshal(edited, &task)
 			if err == nil {
 				break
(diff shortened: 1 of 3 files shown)


naggie:
I think we should run the task name through a whitelist filter and drop any characters not on the list. Also, convert spaces to dashes, and normalise to lowercase. Numbers and a-z should bring enough information over.

dontlaugh:
@naggie Ready for review!

naggie:
Thanks @dontlaugh. Nitpicking --- but if we filter characters that aren't in the whitelist we don't end up with repeated dashes; would you mind changing it so we don't get repeated dashes please?

dontlaugh:
@naggie Fixed, I think. I've updated the test cases. Take a look at those to see the current behavior.

naggie:
Thanks @dontlaugh !

PR #166, opened 2022-03-12
Add show-next to valid command switch statement

### LINKED ISSUE #164: Panic on `dstask show-next`
When executing `dstask show-next`, the following output appears:

```
panic: this should never happen?

goroutine 1 [running]:
main.main()
	./main.go:167 +0x1888
```

I suspect that this is due to the fact that `show-next` is an unhandled command.

dontlaugh:
"show-next" is parsed as a command, but was not being matched on in the
main command switch statement. This caused it to fall through to our
panic statement. Fix this by making it behave the same as "next" or not
passing any argument at all.

Fixes #164

### DIFF
--- a/cmd/dstask/main.go
+++ b/cmd/dstask/main.go
@@ -39,7 +39,7 @@ func main() {
 
 	switch query.Cmd {
 	// The default command
-	case "", dstask.CMD_NEXT:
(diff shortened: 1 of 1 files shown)


naggie:
Thanks @dontlaugh

PR #170, opened 2022-11-06
Compile binaries for Apple silicone

kingo55:
Apple Silicone is being used more and more (including on one of my devices), so I figured we should build and release binaries for those users on M1/M2 chips who don't run Rosetta.

I'm not a go dev, but this is what I've used to compile and run a `dstask` binary on my hardware. I also don't use/haven't tested the `dstask-import` binary on my hardware.


dontlaugh:
Since dstask is pure-Go, cross compilation should work. I haven't tested this yet, though.

dontlaugh:
@naggie Looked over this script. It's up to you, but I'd be inclined to merge this. You can simply note that apple silicon is experimental.

naggie:
Merged! Sorry I forgot about this. I'll spin a release in a few days. Thanks @kingo55 and @dontlaugh

PR #177, opened 2023-03-28
deps: update xurls to v2

khardix:
I'm toying with packaging dstask for Fedora, and this is the one dependency that needed an update 🙂

I have only a very passing knowledge on how to write Go, so please bear with me in case of "obvious" mistakes.


naggie:
Thanks @khardix ! I'll sort out CI and when it passes I'll merge. By the way, I think I will get rid of the vendored code, and stick to go module with remotes to manage dependencies. The diffs are just too unwieldy. If you get a sec perhaps you could do so? I will be away from the computer for a few days.

PR #178, opened 2023-07-17
🐛  Do not parse hidden files like .gitkeep

thatInfrastructureGuy:
- 🐞 Fixes bug where it tries to parse .gitkeep are throws error `error loading task: filename does not encode UUID .gitkeep (wrong length)`
- 🧹 Use os.Readdir instead of deprecate ioutil.Readdir

### DIFF
--- a/task.go
+++ b/task.go
@@ -102,7 +102,7 @@ func (t Task) Equals(t2 Task) bool {
 
 // Unmarshal a Task from disk. We explicitly pass status, because the caller
 // already knows the status, and can override the status declared in yaml.
-func unmarshalTask(path string, finfo os.FileInfo, ids IdsMap, status string) (Task, error) {
+func unmarshalTask(path string, finfo os.DirEntry, ids IdsMap, status string) (Task, error) {
 	if len(finfo.Name()) != TASK_FILENAME_LEN {
 		return Task{}, fmt.Errorf("filename does not encode UUID %s (wrong length)", finfo.Name())
 	}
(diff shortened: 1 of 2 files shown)


go.sum
```diff
@@ -113,8 +113,6 @@ github.com/kr/text v0.1.0 h1:45sCR5RtlFHMR4UwH9sdQ5TC8v0qDQCHnXt+kaKSTVE=
 github.com/kr/text v0.1.0/go.mod h1:4Jbv+DJW3UT/LiOwJeYQe1efqtUx/iVham/4vfdArNI=
 github.com/mattn/go-isatty v0.0.12 h1:wuysRhFDzyxgEmMf5xjvJ2M9dZoWAXNNr5LSBS7uHXY=
 github.com/mattn/go-isatty v0.0.12/go.mod h1:cbi8OIDigv2wuxKPP5vlRcQ1OAZbq2CE4Kysco4FUpU=
-github.com/mvdan/xurls v1.1.0 h1:OpuDelGQ1R1ueQ6sSryzi6P+1RtBpfQHM8fJwlE45ww=
-github.com/mvdan/xurls v1.1.0/go.mod h1:tQlNn3BED8bE/15hnSL2HLkDeLWpNPAwtw7wkEq44oU=
```
naggie, line 117: this seems like an unrelated change

naggie:
Thanks @thatInfrastructureGuy  -- will merge soon

go.sum
thatInfrastructureGuy: Removed go.sum changes.

PR #183, opened 2023-11-30
bugfix: display of line including CJK characters in table

newro:
image

When a task item contains CJK characters such as Hangul,
the table appears to be broken when rendering a list or detail.

To improve this,
I modified the FixStr code to calculate a precisely length using the https://github.com/mattn/go-runewidth.git code.

After fixing
image


naggie:
Useful, thanks @newro

PR #197, opened 2025-03-09
main: do not initialize for version & completion commands

### LINKED ISSUE #196: Remove side effects and dependencies for shell script generation
Hi!

For example, when running `dstask bash-completion`, dstask will first check for git is available in $PATH and then ensure that *~/.dscache* is initialized. This is not necessary because the semantic of this command is to just deliver a static script. This matters because when packaging the application it can be a bit cumbersome to get the right requirements : at least that's the issue I had for NixOS : https://github.com/NixOS/nixpkgs/pull/388299/files#diff-300d67dd6673cd5fd81c8dda79c9fb1c2ee57c78480396c97ed459b4a8bcbed6R30-R31

I'll try to submit a patch tomorrow, but I put this issue here in case I get discouraged while setting up a Go environment 🤷

remi-dupre:
Fixes #196

### DIFF
--- a/cmd/dstask/main.go
+++ b/cmd/dstask/main.go
@@ -10,12 +10,36 @@ import (
 )
 
 func main() {
-	// special case: allow users to run the help command without requiring
-	// initialisation. Other commands are handled in the below switch statement
-	// after initialisation.
-	if len(os.Args) > 1 && os.Args[1] == dstask.CMD_HELP {
+	query := dstask.ParseQuery(os.Args[1:]...)
+
+  // It will remain true if we handle a command that doesn't require
+  // initialisation
+  noInitCommand := true
+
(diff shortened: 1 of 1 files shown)


remi-dupre:
Hi!

That's my first time ever writing anything in go so don't hesitate to push against anything that doesn't seem idiomatic!

(In case anyone wonders, I find the syntax repulsive, but that's only cosmetic and I have to admit it's super good at what it pretends to do : syntax is as simple as C and it compiles instantly)

naggie:
Thanks @remi-dupre much appreciated!

I kind of agree with the syntax etc. You'll get used to it! :-)

Error handling is a pain in the ass in go. However, I've realised as it makes you deal with every error condition then-and-there, your code is much more reliable as a result.

Looks good to me!

Issue #198, opened 2025-04-19
dstask sync "hardcode" master


choman:
I'm starting up again with dstask and noticed that the dstask sync tries to goto master, which unfortunately has been almost completely taking over by main

naggie:
it's possible to detect the branch with `git symbolic-ref refs/remotes/origin/HEAD --short` -- or switch to implicit remote tracking branches. PR welcome!

naggie:
I've found this to be more reliable (in python)

<img width="559" alt="Image" src="https://github.com/user-attachments/assets/11806c7b-4563-4c87-8558-053a4b58d2a1" />

PR #201, opened 2025-05-16
feat: Remove origin remote and master

ylchen07:


naggie:
I guess though, this isn't set when dstask creates the repo as there's no remote. I suggest a message to ensure that a remote is set. 

Is it also clear what the problem is if you try and `sync` without a remote + tracking branch set?

ylchen07:
Related to issue #198

naggie:
Cool merged thanks @ylchen07 !

Issue #211, opened 2025-10-08
Color of Critical priority broken in master


AxelRHD:
Due to the due topic, I installed dstask via go install yesterday, as you suggested. Unfortunately, the colors for the Critical priority are now broken:

<img width="112" height="185" alt="Image" src="https://github.com/user-attachments/assets/926f0726-49d3-42ef-8972-5849e7a83f75" />

It would also be nice if the active tasks had a colored output according to their priority.

But beyond that, I love the tool. It's exactly what I was looking for, as I no longer have to leave my programming shell for task management. :)

naggie:
Good approach, happy to merge :-)

PR #213, opened 2025-10-15
Keep default color for resolved overdue tasks

eduardofuncao:
When running `dstask next`, tasks that are due today or later are colored with FG_PRIORITY_HIGH by default. The same behavior is seen when running `dstask show-resolved`. However, since these tasks are already resolved, it doesn’t make sense to highlight them based on their due dates.

To address this, a simple condition was added to the Task.Style() function: resolved tasks that are past their due dates will no longer be colored.

image

References #211


naggie (approved):

naggie:
Thanks @eduardofuncao

PR #215, opened 2025-10-26
Allow arguments in the EDITOR environment variable.

harkaitz:
Most POSIX utilities such as `crontab -e` or the `git` itself allow
arguments in the EDITOR environment variable. This change makes dstask
behave the same way.

    export EDITOR="emacs -q -nw"
    dstask edit 1

### DIFF
--- a/util.go
+++ b/util.go
@@ -163,10 +163,10 @@ func MakeTempFilename(id int, summary, ext string) string {
 }
 
 func MustEditBytes(data []byte, tmpFilename string) []byte {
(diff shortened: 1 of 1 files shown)


naggie (commented):
Good idea.

I think it would be neater to parse $EDITOR and spread the result with the file name to RunCmd directly -- no need for a function abstraction

harkaitz:
You were right, the code is cleaner that way. I amended the commit.

Issue #216, opened 2025-10-29
show-open sub command displays a truncated task list


IngoMeyer441:
Thanks for `dstask`, it is a great piece of software for all people living in the terminal! 😊 

`dstask help` states that `show-open` displays the task list without truncation, however I get the same truncated task list which is generated by `dstask next`. I digged a bit into the code and the `truncate` paramter [is set to `true` for both commands](https://github.com/naggie/dstask/blob/deaa8729fe1b38cc1c1f7264c9d85524172a3fd1/commands.go#L461):

```go
// CommandShowOpen prints a list of open tasks without truncation.
func CommandShowOpen(conf Config, ctx, query Query) error {
	ts, err := LoadTaskSet(conf.Repo, conf.IDsFile, false)
	if err != nil {
		return err
	}

	query = query.Merge(ctx)
	ts.Filter(query)
	if err := ts.DisplayByNext(ctx, true); err != nil {
		return err
	}

	return nil
}
```

So I guess, it should be

```go
	if err := ts.DisplayByNext(ctx, false); err != nil {
```

?

PR #217, opened 2025-10-31
Fix truncated task view when running `show-open`

IngoMeyer441:
The `show-open` sub command should display the task list without truncation.

This fixes GitHub issue #216.


naggie:
Great thanks @IngoMeyer441

PR #219, opened 2025-11-06
Support getting terminal size on windows

philippbeckmann:
Adds support for getting the terminal size on windows. Tested on Win11 with cmd.exe and Powershell in Windows-Terminal.


naggie:
Seems reasonable, thanks @philippbeckmann

PR #229, opened 2026-04-01
Fix note <id> <text> discarding text without a TTY

niklaas:
`dstask note <id> <text>` silently does nothing when stdout is not a TTY (pipes, scripts, CI, non-interactive shells). The `StdoutIsTTY()` check gates both the interactive editor path and the text-append path, so when there's no TTY the provided text is discarded and the command falls through to printing existing notes instead.

I would like to get the non-interactive flow working so programmatic usage works more easily.

## Fix

Check for provided text first, before the TTY check. Appending text doesn't need a TTY, only the interactive editor does.

```
if query.Text != ""       → append text (no TTY needed)
else if StdoutIsTTY()     → open editor
else                      → print notes to stdout
```

## Additional cleanup

I noticed that the vendor directory was removed from the repo in `bf36524` but `-mod=vendor` flags were left in the build commands. CI worked around this by running `go mod vendor` before builds. This PR drops the stale `-mod=vendor` flags from `integration/main_test.go`, `integrationtest.sh`, and `.github/workflows/ci.yml`, and removes the now-unnecessary `go mod vendor` CI steps.

I only added this change while having trouble to get the integration tests working.

---

Disclaimer: My go knowledge is rather limited and the changes I applied using a coding agent.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,9 +36,6 @@ jobs:
         with:
           go-version-file: go.mod
 
-      - name: Go mod vendor
-        run: go mod vendor
-
       - name: Configure Git User
         run: |
           git config --global user.name "GitHub Actions Bot"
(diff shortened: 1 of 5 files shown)


naggie:
Awesome thanks @niklaas

PR #232, opened 2026-05-02
docs: Fix heading

benasocj:
Put Taskwarrior heading at the same level as the GitHub one.


naggie:
Good catch thanks @benasocj

PR #235, opened 2026-05-11
Add Nushell completions and embed script

cgardner:
Wire `dstask nushell-completion` into the same path as the existing bash/zsh/fish print commands. The new `completions/nushell.nu` script defines a `nu-complete dstask` custom completer that delegates to the existing `dstask _completions` engine, so suggestions for commands, projects, tags, priorities, and templates work the same way they do in other shells.


naggie (approved):
Very cool. Thank you! nushell ftw.

PR #236, opened 2026-06-30
Add fish completion documentation to the readme

AvianAnalyst:
Found this tool today, excited to use it! Had to dig through the issues a bit to find that there was Fish completion support, and how to add it. 

Note: Also mentioned nushell completion in the features list

One thing I wasn't sure of, was if the context awareness for tags mentioned for bash/zsh completion applies to nushell/fish completion as well. If it does, I'm happy to do a quick edit and change the feature list to just list all four.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ Features:
 - `note` command -- edit a **full markdown note** for each task. **Checklists are useful here.**
 - `open` command -- **open URLs found in specified task** (including notes) in the browser
 - zsh/bash completion (including tags and projects in current context) for speed; PowerShell completion on Windows
+- nushell/fish completion
 - A single statically-linked binary
 - [import tool](doc/dstask-import.md) which can import GitHub issues or taskwarrior tasks.
 
@@ -112,6 +113,7 @@ Requirements:
    - Bash: add `source <(dstask bash-completion)` to your `.bashrc`.
    - Zsh: add `dstask zsh-completion > /usr/local/share/zsh/site-functions/_dstask` or source `completions/zsh.sh` in your `~/.zshrc`.
    - Nushell: run `dstask nushell-completion | save -f ~/.config/nushell/completions/dstask.nu` then add `source ~/.config/nushell/completions/dstask.nu` to your `config.nu`.
+   - Fish: add `dstask fish-completion | source` to your `~/.config/fish/config.fish`
    - PowerShell (Windows): see section "PowerShell completion" below.
 1. Set up an alias in your `.bashrc`: `alias task=dstask` or `alias t=dstask` to make task management slightly faster.
 1. Create or clone a `~/.dstask` git repository for the data, if you haven't already:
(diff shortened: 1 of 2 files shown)


naggie:
Thanks! Though I suggest the completions piped to the shell config itself to avoid running a shell command every shell invocation -- that  can quickly build up (with other similar things) and result in a slow shell

AvianAnalyst:
Hi! Open to the suggestion but not sure I understand. My proposed change instructs them to add that line to their shell config, so they wouldn't have to run a shell command every invocation, and it pattern matches to the other instructions. 

Would you prefer that it be a one liner they can run to add it to the config? Something like 
`echo "dstask fish-completion | source" >> ${XDG_CONFIG_HOME}/fish/config.fish` ? I don't think this is the best option as it makes assumptions on the user's configuration setup, and I don't think it is resilient across platforms. 

Or do you mean the actual completion installation is flawed? In that case I'm not sure. I'm pretty new to using fish myself, but this is how all the plugins I've added so far have instructed installation, and I haven't seen any counter examples.

naggie:
All I mean is it is generally better to paste the output of the `dstask <x>-completion` command straight into the file to avoid an unnecessary exe call

AvianAnalyst:
Ahh that makes more sense. And after a little more looking fish auto-picks up completions from configured completions directory ({$XDG_CONFIG_HOME}/fish/completions / ~/.config/fish/completions by default among others). So I'll definitely update the PR, but while looking at it, it looks like the actual fish completion file is also running a command? it has echo (dstask _completions) | string collect. Would it not be better, based on this discussion, to also just have fish-completion output the fully computed completions? 

I'm also wondering if it makes sense to keep zsh-completion, fish-completion and bash-completion IN the completion file. seems like once you have your completion set  up (one time thing per shell, and copy pasteable from the readme) it would just be clutter in your completion options? 

Also also, fish automatically sorts completions alphabetically, but theres a --keep-order switch that disables that. I'm wondering if it would make sense to use that to provide more common commands earlier. At minimum if we don't take out the completion options, they show up quite early. 

Overall what I'd like to do is: 
A) Take the completion commands out of the auto complete file
B) Add --keep-order (-k) to fish_completion output 
C) update fish-completion command to output the completion file in its entirety
`
#!/usr/bin/env fish

complete -f -k -c dstask -a "next add rm remove template log start note notes stop done resolve context modify edit undo sync open git show-next show-projects show-tags showctive show-paused show-open show-resolved show-templates show-unorganised help version P0 P1 P2 P3"
`
This would be the final completion file. (Note, you don't need the commented lines in the current file, fish will still support both abbrs and aliases). 

Examples: 
image
image
image

(You can ignore the G abbr, this is a global abbr I have configured and is not from the dstask completion config). 

If you are on board with these changes, would you like me to wrap them all up into this pr? Or would you like me to edit this pr to be for the current version, and submit a new pr that makes those changes to the completions stuff and updates the documentation accordingly (if need be).

naggie:
Sounds good and well considered, please go ahead thanks

AvianAnalyst:
I'll do that! Would you like it in this pr, or would you like a new pr with those changes?

AvianAnalyst:
Actually looking at it a bit more, fish stores user and system config directories at `$__fish_config_dir` and `$__fish_sysconf_dir` vars. So I could add/edit the fish completion command to instead of outputting the completion file, just save the file to `$__fish_config_dir/completions/dstask.fish` by default, and accept a -S or --system flag to instead save to `$__fish_sysconf_dir/completions/dstask.fish`. 

Thoughts on that?

AvianAnalyst:
The current push is the changes we've discussed. Scratch my last recommendation, after some frustrating debugging trying to implement it, fish doesn't export the config directories. so I've updated the readme to just do a file creation operation from fish where they can use that env variable.

AvianAnalyst:
Hey just wanted to reach out and see if you needed anything else from this?

